PR #3, opened 2019-12-13
Warn about not implementing complex calculations in DataGrid's calculateCellValue

followmyutopia:
https://trello.com/c/GWNNpMlV/1279-calculatecellvalue-describe-specifics-of-this-function

### DIFF
--- a/api-reference/_hidden/GridBaseColumn/calculateCellValue.md
+++ b/api-reference/_hidden/GridBaseColumn/calculateCellValue.md
@@ -365,7 +365,7 @@ To invoke the default behavior, call the **this.defaultCalculateCellValue(rowDat
     value: "column's configuration"
 }
 
-[note] This function is called multiple times for every record: when the widget renders it, applies sorting or filtering, and computes summaries. We recommend that you keep calculations inside this function as simple as possible so as not to lower the widget's performance.
+[note] This function is called multiple times for every record: when the widget renders it, when it applies sorting or filtering, and when it computes summaries. It is recommend that you keep calculations inside this function as simple as possible to avoid hindering widget performance.
 
 #####See Also#####
 - **columns[]**.[customizeText](/api-reference/_hidden/GridBaseColumn/customizeText.md '{basewidgetpath}/Configuration/columns/#customizeText')
--- a/api-reference/_hidden/GridBaseColumn/calculateCellValue.md
+++ b/api-reference/_hidden/GridBaseColumn/calculateCellValue.md
@@ -269,8 +269,6 @@ The following features are disabled in an unbound column, but you can enable the
     </table>
 </div>
 
-[note]
-
 To invoke the default behavior, call the **this.defaultCalculateCellValue(rowData)** function and return its result.
 
 ---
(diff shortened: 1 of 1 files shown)


api-reference/_hidden/GridBaseColumn/calculateCellValue.md
```diff
@@ -363,12 +361,12 @@ To invoke the default behavior, call the **this.defaultCalculateCellValue(rowDat
 
 ---
 
-[/note]
-
 #include uiwidgets-ref-functioncontext with { 
     value: "column's configuration"
 }
 
+[important] Do not implement complex calculations within this method to avoid poor widget's performance. This method is called every time a row renders or when a sorting/filtering/summary setting changes. Thus, it can be called multiple times in a short period and multiple complex calculations can slow down the widget.
```
RomanTsukanov: We warn the reader but do not provide a way to solve the issue. What calculation is considered complex? What to do if the reader still needs this calculation to be performed?

@vconst Can't we fix this in code? Why does this function need to be called when data is sorted, filtered, etc.? Why recalculate the values when they were already calculated?

vconst: Adding cache for such cases is not simple task. And in most of user cases cache is not necessary and it may be bad because cache takes up addition memory.

If user has large calculations in caclulateCellValue, he can add caching to his own code.

So, customers should know that caclulateCellValue is called many times and they should optimize their own code by simplifing or by adding cache.

RomanTsukanov: > If user has large calculations in caclulateCellValue, he can add caching to his own code.

Any good articles about caching to which we could forward the reader?

vconst: I do not have any specific links for this theme

vconst (approved):

RomanTsukanov (approved):

api-reference/_hidden/GridBaseColumn/calculateCellValue.md
```diff
@@ -363,12 +361,12 @@ To invoke the default behavior, call the **this.defaultCalculateCellValue(rowDat
 
 ---
 
-[/note]
-
 #include uiwidgets-ref-functioncontext with { 
     value: "column's configuration"
 }
 
+[note] This function is called multiple times for every record: when the widget renders it, applies sorting or filtering, and computes summaries. We recommend that you keep calculations inside this function as simple as possible so as not to lower the widget's performance.
```
arminal:
```suggestion
[note] This function is called multiple times for every record: when the widget renders it, when it applies sorting or filtering, and when it computes summaries. It is recommend that you keep calculations inside this function as simple as possible to avoid hindering widget performance.
```

RomanTsukanov: > This function is called multiple times for every record: when the widget renders it, ...

Here, "it" refers to the record, while later:

> ... when it applies sorting or filtering, and when it computes summaries.

... "it" refers to the widget.

arminal: can you rewrite it?

RomanTsukanov:
```suggestion
[note] This function is called multiple times for every record: when the record is rendered, when sorting or filtering is applied, and when summaries are computed. It is recommend that you keep calculations inside this function as simple as possible to avoid hindering widget performance.
```

PR #54, opened 2019-12-30
Add Vue and React examples to DataGrid - Enhance Performance and Columns

RomanTsukanov:


concepts/05 Widgets/DataGrid/10 Enhance Performance on Large Datasets/010 Remote Operations.md
```diff
@@ -47,4 +47,67 @@ Specify the [remoteOperations](/api-reference/10%20UI%20Widgets/dxDataGrid/1%20C
...
+    <!-- tab: App.vue -->
+    <template>
+        <DxDataGrid ... >
+            <DxRemoteOperations
+                :filtering="true"
+                :paging="true"
+                :sorting="true"
+                :summary="true"
+                :grouping="true"
+                :groupPaging="true"
```
artem-kurchenko: should be 
:group-paging="true"

concepts/05 Widgets/DataGrid/15 Columns/35 Hide a Column Using the API.md
```diff
@@ -26,6 +26,78 @@ A column is considered hidden when its [visible](/api-reference/_hidden/GridBase
...
+
+    import DataGrid from 'devextreme-react/data-grid';
+
+    class App extends React.Component {
+        constructor(props) {
+            super(props);
+            this.dataGridRef = React.createRef();
+
+            this.hideEmails = () => {
+                this.dataGrid.columnOption('Email', 'visible', false);
```
artem-kurchenko: This approach is not good for React. Let's use React State to change the visible option.

You can check my examples in the Filtering guides (e.g. https://js.devexpress.com/Documentation/Guide/Widgets/DataGrid/Filtering_and_Searching/#Filter_Row)

concepts/05 Widgets/DataGrid/15 Columns/50 Adaptability.md
```diff
@@ -42,6 +42,61 @@ When adapting to a small container or screen, the **DataGrid** can hide columns.
...
+    import DataGrid, {
+        Column
+    } from 'devextreme-react/data-grid';
+
+    class App extends React.Component {
+        render() {
+            return (
+                <DataGrid ...
+                    columnHidingEnabled={true}>
+                    <Column hidingPriority={2} ... /> {/* a valuable column */}
```
artem-kurchenko: I believe it should be defaultHidingPriority here

artem-kurchenko: However, if it's OK it won't be changed in UI. So, we can leave it as is.

concepts/05 Widgets/DataGrid/15 Columns/35 Hide a Column Using the API.md
RomanTsukanov: I decided to use this approach because it is universal if you declare columns in the markup as well as if you use `customizeColumns`.

artem-kurchenko: Ah, I got your idea. Then, it's strange that in Vue we demonstrated two-way binding.
:visible.sync="isEmailVisible"
At least we need to change Vue for consistency. Or (what is better) we can demonstrate two possible solutions (bindings and our API)  for Angular, React, and Vue :)

What do you think?

RomanTsukanov: I've updated the code. It uses two-way binding now.

artem-kurchenko (approved):

AlexJameson (approved):

PR #117, opened 2020-01-17
Draft! ExcelJS - docs

EugeniyKiyashko:


RomanTsukanov:
Great attempt at writing the docs! We can't use this material as is, but we certainly will borrow from it when we get to describing the feature. For now, I'm closing this PR. Thanks!

PR #243, opened 2020-02-03
Fix jQuery snippets in columns.headerFilter.dataSource

onebalaban:


RomanTsukanov (approved):

artem-kurchenko (approved):

artem-kurchenko:
@onebalaban  one more note. Maybe we should add Vue and React here as we're now processing this topic? :)

onebalaban:
> @onebalaban one more note. Maybe we should add Vue and React here as we're now processing this topic? :)

In this PR, I fixed the issue with jQuery only. I will add React and Vue snippets for articles including this one in the context of another task.

PR #297, opened 2020-02-10
Update dataType.md

HumaMilisic:
'guid' is a valid and functional type (angular 7.2.0)


RomanTsukanov:
Hi @HumaMilisic,

Although GUID is indeed a valid type in Angular, DataGrid does not process the GUID values in any special way, and therefore the `dataType` option does not have the `"guid"` value. For that reason, I'm closing this pull request.

PR #331, opened 2020-02-13
Add Vue and React examples for the Calendar

DokaRus:

PR #333, opened 2020-02-13
Add Angular, Vue and React examples for the Switch

DokaRus:
Here, I used a code snippet extension proposed by @artem-kurchenko. Is that a priority way?

### DIFF
--- a/concepts/05 Widgets/Switch/00 Overview.md
+++ b/concepts/05 Widgets/Switch/00 Overview.md
@@ -11,7 +11,7 @@ The following code adds the **Switch** to your page.
 
     <!-- tab: index.js -->
     $(function() {
-        $("#{widgetName}Container").dx{WidgetName}({
+        $("#switchContainer").dxSwitch({
             value: true
         });
     });
@@ -19,8 +19,8 @@ The following code adds the **Switch** to your page.
 ##### Angular
 
     <!-- tab: app.component.html -->
-    <dx-{widget-name} [value]="true">
-    </dx-{widget-name}>
+    <dx-switch [value]="true">
+    </dx-switch>
 
     <!-- tab: app.component.ts -->
     import { Component } from '@angular/core';
@@ -37,7 +37,7 @@ The following code adds the **Switch** to your page.
     import { NgModule } from '@angular/core';
     import { AppComponent } from './app.component';
 
-    import { Dx{WidgetName}Module } from 'devextreme-angular';
+    import { DxSwitchModule } from 'devextreme-angular';
 
     @NgModule({
         declarations: [
@@ -45,7 +45,7 @@ The following code adds the **Switch** to your page.
         ],
         imports: [
             BrowserModule,
-            Dx{WidgetName}Module
+            DxSwitchModule
         ],
         providers: [ ],
         bootstrap: [AppComponent]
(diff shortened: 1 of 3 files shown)


concepts/05 Widgets/Switch/10 Handle the Value Change Event.md
```diff
@@ -1,16 +1,109 @@
...
-    <!--JavaScript-->
+---
+##### jQuery
+
+    <!-- tab: index.js -->
     $(function() {
-        $("#switchContainer").dxSwitch({
-            onValueChanged: function (e) {
+        $("#{widgetName}Container").dx{WidgetName}({
+             onValueChanged: function (e) {
```
AlexJameson:
```suggestion
            onValueChanged: function (e) {
```

concepts/05 Widgets/Switch/20 Keyboard Support.md
```diff
@@ -30,28 +30,36 @@ You can implement a custom handler for a key using the [registerKeyHandler(key,
 ---
 ##### jQuery
 
-    <!--JavaScript-->
+    <!-- tab: index.js -->
     function registerKeyHandlers () {
-        let switch =  $("#switchContainer").dxSwitch("instance");
+        const switch =  $("#switchContainer").dxSwitch("instance");
```
AlexJameson, line 35: Not sure if we should use const in jQuery. Better ask @RomanTsukanov about this

concepts/05 Widgets/Switch/00 Overview.md
```diff
@@ -6,17 +6,93 @@ The **Switch** is a widget that can be in two states: "On" (when [value](/api-re
...
-     
- 
-    <!--JavaScript-->
+---
+##### jQuery
+
+    <!-- tab: index.js -->
     $(function() {
-        $("#switchContainer").dxSwitch({
+        $("#{widgetName}Container").dx{WidgetName}({
```
RomanTsukanov: It is better not use {widgetName} if the file is not inherited. Please replace these constructions with the real widget name across all these documents.

concepts/05 Widgets/Switch/20 Keyboard Support.md
RomanTsukanov: According to [can i use](https://caniuse.com/#search=const), we can use `const` in this particular case.

concepts/05 Widgets/Switch/10 Handle the Value Change Event.md
```diff
@@ -11,6 +14,96 @@ To process a new **Switch** value, you need to handle the value change event. If
         });
     });
 
+##### Angular
+
+    <!-- tab: app.component.html -->
+    <dx-{widget-name} (onValueChanged)="switchValueChanged($event)">
```
RomanTsukanov: It looks like in Angular and Vue we should show two-way binding and in React state changing.

concepts/05 Widgets/Switch/20 Keyboard Support.md
```diff
@@ -61,13 +69,84 @@ You can implement a custom handler for a key using the [registerKeyHandler(key,
...
+        providers: [ ],
+        bootstrap: [AppComponent]
     })
+    export class AppModule { }
+
+##### Vue
+
+    <!-- tab: App.vue -->
+    <template>
+        <Dx{WidgetName} @initialized="registerKeyHandlers" />
```
RomanTsukanov: We usually call `registerKeyHandler` on `mounted`/`componentDidMount()`

RomanTsukanov (changes requested):

concepts/05 Widgets/Switch/10 Handle the Value Change Event.md
DokaRus: [please notice](https://github.com/DevExpress/devextreme-documentation/pull/331#discussion_r380236985)

RomanTsukanov (approved):
👍

PR #477, opened 2020-03-10
Add React and Vue snippets to DataGrid_Sorting_API

onebalaban:


concepts/05 Widgets/DataGrid/25 Sorting/20 API/1 Initial and Runtime Sorting.md
```diff
@@ -73,5 +133,82 @@ Change the **sortOrder** and **sortIndex** options using the [columnOption](/api
...
+        }
+
+        render() {
+            return (
+                <DataGrid ref={this.dataGridRef} ...>
+                </DataGrid>
+            );
+        }
+
+        sortByCountries() {
```
RomanTsukanov: This function should be bound to the component's context.

concepts/05 Widgets/DataGrid/25 Sorting/20 API/5 Custom Sorting.md
```diff
@@ -35,7 +35,58 @@ Implement a custom sorting routine using the [calculateSortValue](/api-reference
...
         // ...
     })
-    
+
+##### Vue
+
+    <!-- tab: App.vue -->
+    <template>
+        <DxDataGrid>
+            <!-- data-field provides values for the column, calculate-sort-value provides values to be used in sorting -->
```
RomanTsukanov:
```suggestion
            <!-- data-field provides values for the column -->
            <!-- calculate-sort-value provides values to be used in sorting -->
```

concepts/05 Widgets/DataGrid/25 Sorting/20 API/5 Custom Sorting.md
```diff
@@ -35,7 +35,58 @@ Implement a custom sorting routine using the [calculateSortValue](/api-reference
...
+    import 'devextreme/dist/css/dx.common.css';
+    import 'devextreme/dist/css/dx.light.css';
+
+    import { DataGrid, Column } from 'devextreme-react/data-grid';
+
+    class App extends React.Component {
+        render() {
+            return (
+                <DataGrid>
+                    {/* dataField provides values for the column, calculateSortValue provides values to be used in sorting */}
```
RomanTsukanov:
```suggestion
                    {/* dataField provides values for the column */}
                    {/* calculateSortValue provides values to be used in sorting */}
```

concepts/05 Widgets/DataGrid/25 Sorting/20 API/5 Custom Sorting.md
```diff
@@ -92,5 +143,92 @@ Implement a custom sorting routine using the [calculateSortValue](/api-reference
...
+            DxDataGrid,
+            DxColumn
+        },
+        data() {
+            return {
+                dataGridRefKey
+            };
+        },
+        methods: {
+            calculateSortValue(rowData) {
```
RomanTsukanov: You can move `calculateSortValue` to `data()`, and the DataGrid instance won't be needed. In this case, `this` will point to the column:

```js
data() {
    return {
        calculateSortValue: function(rowData) {
            if (rowData.Position === 'CEO')
                return this.sortOrder === 'asc' ? 'aaa' : 'zzz'; // CEOs are always displayed at the top  
            else
                return rowData.Position; // Others are sorted as usual
        }
    };
}

```

concepts/05 Widgets/DataGrid/25 Sorting/20 API/5 Custom Sorting.md
```diff
@@ -92,5 +143,92 @@ Implement a custom sorting routine using the [calculateSortValue](/api-reference
...
+                <DataGrid ref={this.dataGridRef} ...>
+                    <Column
+                        calculateSortValue={this.calculateSortValue}
+                        dataField="Position"
+                        sortOrder="asc" />
+                </DataGrid>
+            );
+        }
+
+        calculateSortValue(rowData) {
```
RomanTsukanov, line 206: Same thing should be possible in React.

RomanTsukanov (changes requested):

concepts/05 Widgets/DataGrid/25 Sorting/20 API/1 Initial and Runtime Sorting.md
```diff
@@ -73,5 +133,82 @@ Change the **sortOrder** and **sortIndex** options using the [columnOption](/api
...
+            this.dataGrid.columnOption('Country', {
```
artem-kurchenko: We don't use columnOption for React, let's use React state instead
like here - https://js.devexpress.com/Documentation/Guide/Widgets/DataGrid/Filtering_and_Searching/#Header_Filter
This is more native for React

concepts/05 Widgets/DataGrid/25 Sorting/20 API/1 Initial and Runtime Sorting.md
```diff
@@ -73,5 +133,82 @@ Change the **sortOrder** and **sortIndex** options using the [columnOption](/api
...
+            DxDataGrid
+        },
+        data() {
+            return {
+                dataGridRefKey
+            };
+        },
+        methods: {
+            sortByCountries() {
+                this.dataGrid.columnOption('Country', {
```
artem-kurchenko: use Vue bindings instead of regular API

concepts/05 Widgets/DataGrid/25 Sorting/20 API/8 Clear Sorting Settings.md
```diff
@@ -31,7 +31,85 @@ You can clear sorting settings for all columns by calling the [clearSorting()](/
...
+
+        render() {
+            return (
+                <DataGrid ref={this.dataGridRef} ...>
+                </DataGrid>
+            );
+        }
+
+        clearSortingByNames() {
+            this.dataGrid.columnOption("Name", "sortIndex", undefined);
```
artem-kurchenko, line 126: Same here: use State instead of `columnOption`

concepts/05 Widgets/DataGrid/25 Sorting/20 API/8 Clear Sorting Settings.md
```diff
@@ -31,7 +31,85 @@ You can clear sorting settings for all columns by calling the [clearSorting()](/
...
+            DxDataGrid
+        },
+        data() {
+            return {
+                dataGridRefKey
+            };
+        },
+        methods: {
+            clearSortingByNames() {
+                this.dataGrid.columnOption("Name", "sortIndex", undefined);
```
artem-kurchenko: Vue bindings

concepts/05 Widgets/DataGrid/25 Sorting/20 API/1 Initial and Runtime Sorting.md
```diff
@@ -18,60 +19,215 @@ Rows are sorted according to the data source by default. Set the [sortOrder](/ap
...
+    import 'devextreme/dist/css/dx.light.css';
+
+    import { DataGrid, Column } from 'devextreme-react/data-grid';
+
+    class App extends React.Component {
+        render() {
+            return (
+                <DataGrid>
+                    <Column
+                        sortIndex={1}
```
RomanTsukanov: It may be a good idea to use `defaultSortIndex` in both cases since the sort index may be changed at runtime, and you do not handle this change in your code.

concepts/05 Widgets/DataGrid/25 Sorting/20 API/1 Initial and Runtime Sorting.md
```diff
@@ -18,60 +19,215 @@ Rows are sorted according to the data source by default. Set the [sortOrder](/ap
...
-            this.dataGrid.instance.columnOption("Country", {
-                sortOrder: "desc",
-                sortIndex: 2
-            });
+        sortIndex: number = 0;
+        sortOrder: string = "asc";
+
+        sortByCountries(index: number, order: string) {
+            this.sortIndex = index;
+            this.sortOrder = order;
```
RomanTsukanov: The names of the binding properties should somehow indicate that they are used for the Country column only. I suggest `countrySortIndex` and `countrySortOrder`.

Besides, it is peculiar that the method demands an `index`, which shouldn't be necessary to apply sorting. I would expect that the required index would be somehow calculated by the method.

RomanTsukanov (dismissed):

concepts/05 Widgets/DataGrid/25 Sorting/20 API/1 Initial and Runtime Sorting.md
artem-kurchenko: We still need to handle onOptionChanged in React, as other default sorting changes need to be processed as well

onebalaban: I agree with countrySortIndex and countrySortOrder.  

As for the method, in jQuery we set options via columnOption:
`dataGrid.columnOption("Country", {
        sortOrder: "desc",
        sortIndex: 2
    });`
Above, we wrote: _Change the **sortOrder** and **sortIndex** options_ 
As you can see, we simply set these options at runtime without additional calculations.

In Angular/React/Vue, we decided not to use a widget instance and columnOption, so I added a simple method that performs the same operation when it is invoked.

If the method signature may mislead the customer, we can simplify it:
`sortByCountries()
            this.countrySortIndex = 2;
            this.countrySortOrder = 'desc';
        }`

RomanTsukanov: Let's change the sortOrder only - the last unocuppied sortIndex will be set automatically.

concepts/05 Widgets/DataGrid/25 Sorting/20 API/1 Initial and Runtime Sorting.md
```diff
@@ -18,60 +19,214 @@ Rows are sorted according to the data source by default. Set the [sortOrder](/ap
...
+            this.setState({
+                countrySortOrder: order
+            });
+        }
+
+        onOptionChanged = (e) => {
+            if (e.fullName === "columns[0].sortOrder") {
+                this.setState({
+                    countrySortOrder: e.value
+                });
```
RomanTsukanov:
```suggestion
                this.sortByCountries(e.value);
```

artem-kurchenko (approved):

PR #524, opened 2020-03-18
Enhance the description of the allowReordering option

AlexJameson:


api-reference/10 UI Widgets/GridBase/1 Configuration/rowDragging/allowReordering.md
```diff
@@ -8,4 +8,8 @@ default: false
 Allows row reordering using drag and drop gestures.
 
 ---
-<!-- Description goes here -->
+This option only enables dragging. You should also implement the reordering logic in the [onReorder]({basewidgetpath}/Configuration/rowDragging/#onReorder) function. If you need to perform cross-component drag-and-drop, implement the reordering logic in the [onAdd]({basewidgetpath}/Configuration/rowDragging/#onAdd) handler.
```
RomanTsukanov: It's unclear whether I should implement both these functions if I need reordering _and_ cross-component drag-and-drop or only `onAdd` will do.

RomanTsukanov: Both `onReorder` and `onAdd` are functions and handlers, but when you call one of them a "function" and the other a "handler", it seems that they are from different classes.

RomanTsukanov (changes requested):

RomanTsukanov (approved):

dirk-pieterse (approved):

PR #551, opened 2020-03-24
broken URL

foaudkajj:
Online Devextreme Theme Builder is broken.


RomanTsukanov:
Hi @foaudkajj,

This PR fixes the link in the documentation for DevExtreme v16.2, but we only change documentation for two most recent DevExtreme versions. As of today, these are v19.1 and v19.2, and they do not contain this broken link because the article about the ThemeBuilder has been reworked. I'm going to close this PR.

PR #671, opened 2020-04-20
Fix T876719 -  It's unclear why nested events should be handled using square brackets

RomanTsukanov:


artem-kurchenko (approved):
ok, let's try)

PR #775, opened 2020-05-18
Fix mentions of the old export in Charts

AlexJameson:


dxBeardedBear (approved):

alexanderPolosatov (approved):

RomanTsukanov (changes requested):
Server-side proxy is no longer required. I think we should remove the guides alrogether.

RomanTsukanov (approved):

PR #875, opened 2020-06-05
DX React Template: Describe authentication

RomanTsukanov:


tar-g (approved):

concepts/50 React Components/02 Create a DevExtreme Application/45 Authentication/00 Authentication.md
```diff
@@ -0,0 +1,5 @@
+A DevExtreme React application includes authentication UI and API. Based on the user's authentication status, the application can be divided into two parts: public pages and protected pages.
+
+Unauthenticated users can access only public pages. These pages provide authentication functionality (sign in, create a new account, reset password). They are rendered by the `NonAuthenticatedContent.js` component.
+
+Protected pages include the rest of the application and can be viewed by authenticated users only. These pages are rendered by the `Content.js` component.
```
IlyaKhD: I suggest to rephrase this part to avoid misunderstanding about what is secure and who is responsible for security.

```suggestion
 Routing is configured to display only authentication pages if a user identity is not available (e.g. a user is not authenticated on the server). These pages allows a user to sign in, to create a new account or to reset password. They are rendered by the `NonAuthenticatedContent.js` component. The rest of the application is rendered by the `Content.js` component.

Note that client-side routing is not protected from malefactors so you should check authentication on back-end.
```

IlyaKhD: @RomanTsukanov, @tar-z what do you think about it?

IlyaKhD (approved):

dirk-pieterse (approved):

PR #944, opened 2020-06-16
Adding accepted values

Hypenate:


AlexJameson:
Hello Ben, thank you for taking the time to submit this correction!

The `Boolean` type indicates the possible values and we consider it sufficient.

PR #1159, opened 2020-08-14
Inject video into Vue App Template documentation

MaximKudriavtsev:


RomanTsukanov (commented):
We agreed in our Teams discussion that Angular and Vue articles should retain images.

PR #1205, opened 2020-08-28
Update direction.md

superole:
The short description was misleading at best. I corrected it to say the same as the text in the demo, only slightly more specific.


AlexJameson:
Hello @superole, thank you for you suggestion. We decided to partially merge your pull request to make the description more clear, but not to copy the text from the demo.

PR #1414, opened 2020-10-20
utils: description for getTimeZones and time zone object

pavel-2020:


api-reference/50 Common/Object Structures/dxSchedulerTimeZone/dxSchedulerTimeZone.md
```diff
@@ -3,7 +3,7 @@ id: dxSchedulerTimeZone
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+A time zone object.
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+Objects of this type constitute the array returned by the [getTimeZones](/Documentation/ApiReference/Common/Utils/utils/#getTimeZonesdate) utility method.
```
RomanTsukanov: We may use this object in other places in the future.

```suggestion
<!-- Description goes here -->
```

api-reference/50 Common/Object Structures/dxSchedulerTimeZone/id.md
```diff
@@ -4,7 +4,7 @@ type: String
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+A time zone text string from the IANA database.
```
RomanTsukanov: Let's add a link to the database.

api-reference/50 Common/utils/utils/getTimeZones(date).md
```diff
@@ -5,13 +5,26 @@ export: getTimeZones
...
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+The following code illustrates how to use this method:
+
+    const timeZones = DevExpress.utils.getTimeZones(new Date(2020, 6, 21));
+    // ===== or when using modules =====
+    import timeZoneUtils from 'devextreme/time_zone_utils';
+    const timeZones = timeZoneUtils.getTimeZones(new Date(2020, 6, 21));
```
RomanTsukanov, line 24:
```suggestion
    import { getTimeZones } from 'devextreme/utils';
    const timeZones = getTimeZones(new Date(2020, 6, 21));
```

api-reference/50 Common/utils/utils/getTimeZones(date).md
```diff
@@ -5,13 +5,26 @@ export: getTimeZones
...
+
+The returned array consists of objects in the following format:
+
+    { offset: 2, title: "(GMT +02:00) Europe/Berlin", id: "Europe/Berlin" }
```
RomanTsukanov: It looked better when it was a comment.

RomanTsukanov (approved):

api-reference/50 Common/Object Structures/dxSchedulerTimeZone/title.md
```diff
@@ -4,7 +4,7 @@ type: String
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+A formatted time zone representation. Includes the **id** and **offset** values.
```
dirk-pieterse: Why 'representation'?

api-reference/50 Common/utils/utils/getTimeZones(date).md
```diff
@@ -5,13 +5,28 @@ export: getTimeZones
...
-<!-- Description goes here -->
+Gets the list of <a href="https://en.wikipedia.org/wiki/List_of_tz_database_time_zones" target="_blank">IANA</a> time zone objects.
 
 ##### return: Array<dxSchedulerTimeZone>
-<!-- Description goes here -->
+A list of IANA time zone objects.
 
 ##### param(date): Date | undefined
-<!-- Description goes here -->
+A date for which to get time zones. If not passed, the current date is used.
```
dirk-pieterse: "A date for which to get time zones" This does not make sense. Time zones are the same regardless of the date.

dirk-pieterse (approved):

api-reference/50 Common/Object Structures/dxSchedulerTimeZone/title.md
pavel-2020: It's not a time zone, but a formatted string that contains a time zone id and its offset. E.g.: title: "(GMT +02:00) Europe/Berlin"

api-reference/50 Common/utils/utils/getTimeZones(date).md
pavel-2020: No, they may change due to daylight savings or other reasons. I've added 2 sentences explaining why this method can accept a specific date.

dirk-pieterse: Time zones do not change - the offset changes with daylight saving time, but the time zone stays the same. Countries can, on rare occasions, change their time zone. The last country to change its time zone was in 1995 if I remember correctly.

api-reference/50 Common/Object Structures/dxSchedulerTimeZone/title.md
dirk-pieterse: "Time zones are often represented by alphabetic abbreviations such as "EST", "WST", and "CST"

"Offsets from UTC/GMT are written in the format ±[hh]:[mm], ±[hh] [mm], or ±[hh] (either hours ahead or behind UTC/GMT). For example, if the time being described is one hour ahead of UTC (such as the time in Berlin during the winter), the zone designator would be "+01:00", "+0100", or simply "+01"."

It does not seem correct to call it "A formatted time zone representation" because 'a formatted representation' can mean different things.

dirk-pieterse (approved):

PR #1496, opened 2020-11-20
Scheduler: recurrenceRuleExpr - remove the line about setting the option to null

pavel-2020:
This feature is extra. To display the initial appointment only, users have to remove a recurrence rule from the appointment or use any other solution.


AlexJameson:
> 
> 
> This feature is extra. To display the initial appointment only, users have to remove a recurrence rule from the appointment or use any other solution.

I'm not an expert in Scheduler, but it seems like this option allows users to disable the recurrence rule for all appointments. This makes it easy for a user to change this option in runtime (for example) when he decides to do it for some reason.

You say that users can remove rules from certain appointments - but most of our widgets allow users to implement common settings as well as the individual ones. 

Or maybe I am missing something?

AlexJameson (approved):

andrewmakarov (approved):

pavel-2020:
The decision to remove this line was made during a daily meeting.

PR #1499, opened 2020-11-21
Add a note to the Declare Content in Markup topic for react

AlexJameson:


concepts/50 React Components/40 Component Configuration Syntax/45 Declare Content in the Markup.md
```diff
@@ -16,12 +16,34 @@ The following is an example with **ScrollView**:
...
+    }
+
+[important]
+
+These widgets do not support dynamically or conditionally rendered content in their root element. For example, the following code **does not work**:
+
+    <!-- tab: App.js -->
+    <React.Fragment>
+        <Drawer ... >
+            <RouterView></RouterView>
```
LazyLahtak: In case of `Vue` the `router-view` is a special `Vue` component, that dynamically re-renders in some case.
For React, we have no such element. I suggest that we can replace this with conditionally contend. For example:
```js
{
  items.map((item, index) => <div key={index}>{item.name}</div>)
}
```
(We can replace `div` by custom element with the `item` property)

LazyLahtak (approved):

SelyankinFyodor (approved):

PR #1546, opened 2020-12-03
20_2 Scheduler: edit the cancel field for onAppointmentFormOpening, onAppointmentClick, onAppointmentDblClick

pavel-2020:
The current versions of onAppointmentClick and onAppointmentDblClick are more precise.
The wording of cancel for onAppointmentFormOpening is improved.

### DIFF
--- a/api-reference/10 UI Widgets/dxScheduler/1 Configuration/onAppointmentClick.md
+++ b/api-reference/10 UI Widgets/dxScheduler/1 Configuration/onAppointmentClick.md
@@ -18,7 +18,7 @@ The initial appointment.
 #include common-ref-elementparam with { element: "clicked" }
 
 ##### field(e.cancel): Boolean
-Allows you to cancel execution of the default appointment click handler.
+If true, hides the appointment tooltip.
 
 ##### field(e.component): {WidgetName}
 The widget's instance.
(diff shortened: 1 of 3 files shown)


api-reference/10 UI Widgets/dxScheduler/1 Configuration/onAppointmentClick.md
```diff
@@ -18,7 +18,7 @@ The initial appointment.
 #include common-ref-elementparam with { element: "clicked" }
 
 ##### field(e.cancel): Boolean
-Allows you to cancel execution of the default appointment click handler.
+If true, hides the appointment tooltip.
```
MaximKudriavtsev:
```suggestion
If **true**, hides the appointment tooltip.
```

Make similar with other

api-reference/10 UI Widgets/dxScheduler/1 Configuration/onAppointmentDblClick.md
```diff
@@ -18,7 +18,7 @@ The initial appointment.
 #include common-ref-elementparam with { element: "clicked" }
 
 ##### field(e.cancel): Boolean
-Allows you to cancel execution of the default appointment click handler.
+If **true**, prevents the appointment details form from opening on double click.
```
AlexJameson, line 21: I would try to avoid 'form from opening'. It is not a rule, but in my opinion in this particular case these words look a bit odd together. 

How about 'If **true**, prohibits the appointment details form opening on double click'?

PR #1606, opened 2020-12-21
Clarify DateBox - disabledDates rules for different views

czernous-devexpress:


api-reference/10 UI Widgets/dxDateBox/1 Configuration/disabledDates.md
```diff
@@ -142,7 +142,8 @@ The current view: *"month"*, *"year"*, *"decade"*, or *"century"*.
 
 ---
 
-... or a function that specifies whether the currently checked date is disabled.
+... or a function that defines rules controlling whether the currently checked date is disabled.
+Make sure to verify that these rules are applied to the desired view.
```
RomanTsukanov: These two lines will be displayed as a single line. Let's make it a single line in the document too.

RomanTsukanov (approved):

api-reference/10 UI Widgets/dxDateBox/1 Configuration/disabledDates.md
```diff
@@ -142,7 +142,7 @@ The current view: *"month"*, *"year"*, *"decade"*, or *"century"*.
 
 ---
 
-... or a function that specifies whether the currently checked date is disabled.
+... or a function that defines rules controlling whether the currently checked date is disabled. Make sure to verify that these rules are applied to the desired view.
```
dirk-pieterse:
```suggestion
... or a function that defines the rules that specify whether the checked date is disabled. You should verify that these rules are applied to a specific view.
```

dirk-pieterse: "Make sure to verify that these rules are applied to the desired view" I'm not entirely sure what this purpose is of this statement. What does 'desired' refer to?

dirk-pieterse (approved):

api-reference/10 UI Widgets/dxDateBox/1 Configuration/disabledDates.md
czernous-devexpress: The purpose of this statement is to indicate that every view should be individually targeted by a separate set of rules. As far as 'desired' goes, it refers to a view that the user intends to target by the rules. Perhaps 'targeted view' as a better alternative.
'Specify' is probably not the best word, as it was used earlier in the text https://js.devexpress.com/Documentation/ApiReference/UI_Widgets/dxDateBox/Configuration/#disabledDates

"Specifies dates that users cannot select." and "... or a function that specifies..."

Could you also please clarify rephrasing "Make sure to..." as "You should..."?

dirk-pieterse: - You can use 'specify' more than once in a doc - the point is not to overuse it, or use it in the same or subsequent sentences thereby creating a pattern
-  'Make sure to' is an instruction for the reader, you can rephrase this as 'You should do something' since you are telling the reader to do something
- 'Make sure to verify that these rules are applied to the desired view' and 'every view should be individually targeted by a separate set of rules' are not the same thing. You should consider rephrasing this sentence

api-reference/10 UI Widgets/dxDateBox/1 Configuration/disabledDates.md
```diff
@@ -142,7 +142,7 @@ The current view: *"month"*, *"year"*, *"decade"*, or *"century"*.
 
 ---
 
-... or a function that specifies whether the currently checked date is disabled.
+Alternatively, pass a function to disabledDates. This function should define the rules that determine whether the checked date is disabled. A separate set of rules should target every [view](/Documentation/ApiReference/UI_Widgets/dxScheduler/Configuration/views/) individually.
```
RomanTsukanov:
```suggestion
Alternatively, pass a function to **disabledDates**. This function should define the rules that determine whether the checked date is disabled. A separate set of rules should target every [view](/Documentation/ApiReference/UI_Widgets/dxScheduler/Configuration/views/) individually.
```

RomanTsukanov (approved):

PR #1746, opened 2021-01-21
Fix DocIssues: Form - Item Options

AlexJameson:


concepts/05 UI Components/Form/20 Change Options at Runtime/10 Item Options.md
```diff
@@ -1,5 +1,7 @@
 To change a single item property at runtime, call the [itemOption(id, option, value)](/api-reference/10%20UI%20Widgets/dxForm/3%20Methods/itemOption(id_option_value).md '/Documentation/ApiReference/UI_Components/dxForm/Methods/#itemOptionid_option_value') method.  This approach is more typical of jQuery. If the needed item is [in a group](/concepts/05%20Widgets/Form/10%20Organize%20Simple%20Items/05%20In%20Groups '/Documentation/Guide/UI_Components/Form/Organize_Simple_Items/In_Groups/') or [in a tab](/concepts/05%20Widgets/Form/10%20Organize%20Simple%20Items/10%20In%20Tabs '/Documentation/Guide/UI_Components/Form/Organize_Simple_Items/In_Tabs/'), the *field* parameter should be given the group caption or tab title followed by the item's name. An example is shown below.
 
+#include form-item-options
```
RomanTsukanov: This note breaks the connection between the text and the code sample. Let's move it to the end of the article.

RomanTsukanov (changes requested):

RomanTsukanov (approved):

includes/form-item-options.md
```diff
@@ -0,0 +1 @@
+When you modify an item, the Form also refreshes all other items in its group. If an item is not in a group, the whole Form is refreshed. To update only the items of your choice, wrap them into a separate group.
```
arminal:
```suggestion
When you modify an item, the Form also refreshes all other items in its group. If an item is not in a group, the whole Form is refreshed. To update only your chosen items, wrap them into a separate group.
```

arminal (commented):
I couldn't edit these sentences in the program, so I am leaving a note here. Can you rewrite these 2? "being called" and "with ***".
Being called with the field* parameter only, this method returns the current configuration of the specified form item.
With Angular, Vue, or React, bind the property that should be changed to a component property.

PR #1889, opened 2021-02-15
Describe the useItemTextAsTitle property in DDList and DDButton

czernous-devexpress:


api-reference/10 UI Components/dxDropDownList/1 Configuration/useItemTextAsTitle.md
```diff
@@ -5,7 +5,7 @@ default: false
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies whether the **{WidgetName}** uses item's text a **title** attribute.
```
RomanTsukanov: Let's add a link to title:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title

api-reference/10 UI Components/dxDropDownList/1 Configuration/useItemTextAsTitle.md
```diff
@@ -5,7 +5,7 @@ default: false
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies whether the **{WidgetName}** uses item's text a [title](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) attribute.
```
RomanTsukanov: It is better to open links to 3rd-party resources in a new tab.

```suggestion
Specifies whether the **{WidgetName}** uses item's text a <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title" target="_blank">title</a> attribute.
```

RomanTsukanov (approved):

PR #2180, opened 2021-04-14
Fix T970871 and remove encodeHtml from docs

AlexJameson:

PR #2260, opened 2021-05-05
Fix T970871 and remove encodeHtml from docs (#2180)

AlexJameson:


concepts/Common/Security Considerations/20 HTML Encoding/10 Encode User Input.md
```diff
@@ -0,0 +1,178 @@
...
+            e.preventDefault();
+        }, []);
+
+
+        return (
+            <form
+                action="your-action"
+                onSubmit={onFormSubmit}
+                <HtmlEditor ... 
+                    value={editorValue}
```
RomanTsukanov, line 168: I'm afraid we forgot `onValueChange`. Without it, `editorValue` won't ever be updated.

RomanTsukanov (changes requested):

concepts/Common/Security Considerations/20 HTML Encoding/10 Encode User Input.md
AlexJameson: I did not forget it, I excluded it on purpose =) Do we need to make this change in all versions in React approach?

RomanTsukanov: Sure

RomanTsukanov (approved):

PR #2346, opened 2021-05-27
Fix headerFilter docissues T884624, T896965, T942522, T855042

RomanTsukanov:
https://supportcenter.devexpress.com/internal/ticket/details/T884624
https://supportcenter.devexpress.com/internal/ticket/details/T942522
https://supportcenter.devexpress.com/internal/ticket/details/T942522
https://supportcenter.devexpress.com/internal/ticket/details/T855042

### DIFF
--- a/api-reference/_hidden/GridBaseColumn/headerFilter/groupInterval.md
+++ b/api-reference/_hidden/GridBaseColumn/headerFilter/groupInterval.md
@@ -15,7 +15,7 @@ For numeric columns, assign a number to this property. This number designates a 
     url: "https://js.devexpress.com/Demos/WidgetsGallery/Demo/DataGrid/Filtering/"
 }
 
-For date columns, set this property to one of the accepted string values above. Dates are grouped into a hierarchy, and the string value indicates its lowest level. The default level is *"day"*, which means that the header filter forms the following hierarchy: Year &rarr; Month &rarr; Day. You can disable the hierarchical display if you set the **groupInterval** to **null**. In this case, you also need to implement the column's [calculateFilterExpression](/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/columns/#calculateFilterExpression) function as follows:
+For date columns, set this property to one of the accepted string values above. Dates are grouped into a hierarchy, and the string value indicates its lowest level. The default level is *"day"* which means that the header filter forms the following hierarchy: Year &rarr; Month &rarr; Day. You can disable the hierarchical display if you set the **groupInterval** to **null**. In this case, you also need to implement the column's [calculateFilterExpression](/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/columns/#calculateFilterExpression) function as follows:
 
 ---
 ##### jQuery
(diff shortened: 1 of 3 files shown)


api-reference/_hidden/GridBaseColumn/headerFilter/groupInterval.md
```diff
@@ -6,12 +6,165 @@ default: undefined
...
+        $("#{widgetName}Container").dx{WidgetName}({
+            // ...
+            columns: [{
+                // ...
+                headerFilter: {
+                    groupInterval: null
+                },
+                calculateFilterExpression(value, operation, target) {
+                    if(value && target === "headerFilter") {
+                        return [this.dataField, operation, new Date(value)];
```
vconst: I tried to copy this example to demo: https://codepen.io/vconst/pen/vYxpOyo. And, It is not necessary to trap value with new Date(). This line can be changed to:
`return [this.dataField, operation, value];`

concepts/05 UI Components/DataGrid/99 How To/Customize Header Filter Data Source/20 Map Data Source Fields.md
```diff
@@ -31,7 +31,9 @@ In the following code, the `categoryName` and `categoryId` fields are mapped to
                                 categoryId: item.categoryId,
                                 categoryCode: item.categoryCode
                             }
-                        }
+                        },
+                        // Turn pagination off if header filter items are sorted incorrectly:
```
vconst: In the original doc issue T818801 postProcess is used to sort items. In this case paginate: false is required. But in all other cases paginate: false do not change sorting. I think we should not add this very specific comment here. Or add some information to DataSource.postProcess topic.

api-reference/_hidden/GridBaseColumn/headerFilter/groupInterval.md
```diff
@@ -6,12 +6,165 @@ default: undefined
...
+For numeric columns, assign a number to this property. This number designates a step with which to generate groups. Column values are classified into these groups.
 
-For numeric columns, assign a number to this property. This number designates a step with which groups should be generated. Column values are classified into these groups.
+#include common-demobutton with {
+    url: "https://js.devexpress.com/Demos/WidgetsGallery/Demo/DataGrid/Filtering/"
+}
+
+For date columns, set this property to one of the accepted string values above. Dates are grouped into a hierarchy, and the string value indicates its lowest level. The default level is *"day"*, which means that the header filter forms the following hierarchy: Year &rarr; Month &rarr; Day. You can disable the hierarchical display if you set the **groupInterval** to **null**. In this case, you also need to implement the column's [calculateFilterExpression](/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/columns/#calculateFilterExpression) function as follows:
```
dirk-pieterse:
```suggestion
For date columns, set this property to one of the accepted string values above. Dates are grouped into a hierarchy, and the string value indicates its lowest level. The default level is *"day"* which means that the header filter forms the following hierarchy: Year &rarr; Month &rarr; Day. You can disable the hierarchical display if you set the **groupInterval** to **null**. In this case, you also need to implement the column's [calculateFilterExpression](/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/columns/#calculateFilterExpression) function as follows:
```

dirk-pieterse (approved):

PR #2487, opened 2021-06-26
Fix T1009709: DataGrid - allowSearch MUST be false if a mapped dataSource is used for a column header filter

RomanTsukanov:


onebalaban (approved):
Great!

PR #2855, opened 2021-09-14
Describe dxActionSheetItem.stylingMode

AlexJameson:


api-reference/_hidden/dxActionSheetItem/stylingMode.md
```diff
@@ -6,7 +6,12 @@ default: 'outlined'
...
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+The following styles are available: 
+
+![ActionSheet Item Styling Modes](/images/UiWidgets/actionSheetItem_stylingMode.png)
+
+#####See Also#####
+- [type](/Documentation/ApiReference/UI_Components/dxActionSheet/Configuration/items/#type)
```
babich-a, line 17: stylungMode?

babich-a (changes requested):

babich-a (approved):

api-reference/_hidden/dxActionSheetItem/stylingMode.md
RomanTsukanov: Backlink from `type` to `stylingMode` maybe?

AlexJameson: Good point. Will add the same change to other members.

DokaRus (approved):

api-reference/_hidden/dxActionSheetItem/stylingMode.md
RomanTsukanov: In case dxActionSheetItem will be inherited by other components:

```suggestion
- [type]({currentpath}/#type)
```

api-reference/_hidden/dxActionSheetItem/type.md
```diff
@@ -9,4 +9,6 @@ default: 'normal'
 Specifies the [type](/api-reference/10%20UI%20Components/dxButton/1%20Configuration/type.md '/Documentation/ApiReference/UI_Components/dxButton/Configuration/#type') of the button that represents an action sheet item.
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+
+#####See Also#####
+- [stylingMode](/Documentation/ApiReference/UI_Components/dxActionSheet/Configuration/items/#stylingMode)
```
RomanTsukanov:
```suggestion
- [stylingMode]({currentpath}/#stylingMode)
```

RomanTsukanov (approved):

babich-a (commented):
Should not we say that this is the same option as dxButton.stylingMode?

api-reference/_hidden/dxActionSheetItem/type.md
```diff
@@ -9,4 +9,6 @@ default: 'normal'
 Specifies the [type](/api-reference/10%20UI%20Components/dxButton/1%20Configuration/type.md '/Documentation/ApiReference/UI_Components/dxButton/Configuration/#type') of the button that represents an action sheet item.
```
albertov05: I'd like to avoid using "represents" if possible. Can we change it to "is"?

AlexJameson:
> 
> 
> Should not we say that this is the same option as dxButton.stylingMode?

I don't think so. It is a separate property of an item (thta is in fact dxButton of course). But we do not declare dxButton explicitly so there is no need to confuse users (IMHO).

api-reference/_hidden/dxActionSheetItem/type.md
AlexJameson: Ok

albertov05: Thanks. I added the suggestion below:

```suggestion
Specifies the [type](/api-reference/10%20UI%20Components/dxButton/1%20Configuration/type.md '/Documentation/ApiReference/UI_Components/dxButton/Configuration/#type') of the button that is an action sheet item.
```

PR #2941, opened 2021-09-30
DataGrid/TreeList.toolbar: Describe empty topics & make some fixes

RomanTsukanov:


api-reference/_hidden/dxTreeListToolbar/disabled.md
```diff
@@ -0,0 +1,11 @@
+---
+id: dxTreeListToolbar.disabled
+type: Boolean
+default: undefined
```
pomahtri:
```suggestion
default: false
```

So it will be the same as in DataGrid's `disabled`

pomahtri (approved):

PR #2953, opened 2021-10-05
Grids: Describe newRowPosition, insertAfter/BeforeKey

RomanTsukanov:


pomahtri:
Do we need to add descriptions of each option of `newRowPosition`? Or it's better not to duplicate it from demo description?

Also as far as I remember, there was some restrictions with `scrolling.mode: 'infinite'`, maybe we need to describe them too. It's question to @Alyar666 or @vconst

vconst (approved):

arminal:
#GrammarReview

Very short doc. A few sentences. No errors.

PR #3031, opened 2021-10-20
DxDiagram - add expression description

LanaDX:


api-reference/10 UI Components/dxDiagram/1 Configuration/edges/textExpr.md
```diff
@@ -14,7 +14,7 @@ The current edge's data object.
 An edge's text.
```
albertov05, line 15: Can you update line 8 (above)?
Line 8:
`Specifies the name of a data source field or an expression that returns edge texts.`

Please change "texts" to "text".

Also, for my understanding, what is an "edge" (and "edge text") in this context?

api-reference/10 UI Components/dxDiagram/1 Configuration/edges/textExpr.md
```diff
@@ -14,7 +14,7 @@ The current edge's data object.
...
 
 ##### param(value): any
-<!-- Description goes here -->
+When the function is called as a setter, returns the edge's new text value; when the function is called as a getter, returns `undefined`.
 
 ---
 The specified data source field or an expression should return a string or object value.
```
albertov05: 1) `The specified data source field or an expression should return a string or object value.`

Is the "expression" something that is also specified? If so, do you mean:

`The specified data source field or expression should return a string or object value.`

2) I'd also like to comment on a few lines below:

Line 22:
`- A string specifies the text displayed in the middle of a connector.`

Please change "A string" to "The string".

Line 24:
`- An object can contain multiple texts and their positions on the connector. The position is a number from 0 to 1, where 0 corresponds to the connector's start point and 1 to the connector's end point.`

Please change "An object" to "The object".

What do you mean by "multiple texts"?

LanaDX: fixed. See my answer above

api-reference/10 UI Components/dxDiagram/1 Configuration/edges/textExpr.md
LanaDX: We name Diagram items in two ways:
- we use `shape` and `connector` words when write about visual elements, displayed on the screen.
- we use `node` and `edge` words when write about data elements, contained in a data source. 
During the binding process, the Diagram creates a shape for every bound node and a connector for every bound edge. 
`edge text` - is a string that belongs to this edge and will be displayed above the connector. 

> What do you mean by "multiple texts"?

A connector can display several texts (for instance, at the start and at the end).  In this case, the textExpr returns an object that contains the texts and their coordinates relatively to the connector.

albertov05: Thank you for the detailed explanation.

AnnaKondratova (approved):

PR #3111, opened 2021-11-18
Fix T1045480 - async/await examples in onRowInserting and similar handlers

RomanTsukanov:


onebalaban (commented):
OK

onebalaban:
Thx. We also need to fix the same in _onRowInserting_ and _onRowUpdating_

onebalaban (approved):
LGTM

PR #3321, opened 2022-01-12
Actualize Gantt columns API

AnnaKondratova:


AnnaKondratova:
This PR updates info in API members inherited from the dxTreeListColumn object.

RomanTsukanov (approved):

PR #3398, opened 2022-01-31
T1011771: add example for custom button type

vladaskorohodova:


AnnaKondratova (approved):

api-reference/10 UI Components/dxButton/1 Configuration/type.md
```diff
@@ -13,6 +13,77 @@ DevExtreme provides the following button types.
...
+    export default App;
+
+    <!--tab: index.css-->
+    .dx-button.dx-button-warning {
+        background-color: #ffc107;
+    }
+
+---
+
+Note that buttons get white background when you move them into the [Toolbar](/Documentation/ApiReference/UI_Components/dxToolbar/). We recommend that you use the *'normal'* type for such buttons, otherwise the button text is not visible, or define custom CSS rules to override the initial styles.
```
dirk-pieterse:
```suggestion
Note that buttons have a white background when you move them to the [Toolbar](/Documentation/ApiReference/UI_Components/dxToolbar/). We recommend that you use the *'normal'* type for such buttons, otherwise the button text is not visible. You can also define custom CSS rules to override the initial styles.
```

dirk-pieterse (approved):

PR #3423, opened 2022-02-07
CustomStore: add example to LoadOptions.userData 

AnnaKondratova:


api-reference/30 Data Layer/CustomStore/LoadOptions/userData.md
```diff
@@ -7,3 +7,122 @@ type: any
...
+---
+##### jQuery
+
+    <!-- tab: index.js -->
+    $(function() {
+        var customDataSource = new DevExpress.data.CustomStore({
+            key: "ID",
+            load: function (loadOptions) {
+                if (!loadOptions.userData["someValue"]) {  
+                    loadOptions.userData["someValue"] = 5;
```
dxBeardedBear: Better to make like this:
let value = loadOptions.userData?.someValue;
if(!value) {
    value = 5;
    alert("Initial value is set");
} else {
    alert("Cached value: " + value);
}

AnnaKondratova: applied

PR #3471, opened 2022-02-25
Remove generate-modules-meta

Igggr:
Command generate-modules-meta was removed in devextreme-internal-tools version 8. 

Before merge - need to merge PR that bump devextreme-internal-tools in DevExtreme, then publish new DevExtreme version and finally bum DevExtreme version here

### DIFF
--- a/.github/workflows/docgen.yml
+++ b/.github/workflows/docgen.yml
@@ -35,10 +35,6 @@ jobs:
           run: npm i
 
 #Generation
-        - name: Discover ts declarations
-          working-direcory: devextreme
-          run: npm run discover-ts-declarations -- --artifacts ../artifacts --js-scripts ../devextreme/js
-
         - name: Discover declarations
           working-directory: devextreme
           run: npm run discover-declarations -- --artifacts ../artifacts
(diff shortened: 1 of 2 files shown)


.github/workflows/docgen.yml
```diff
@@ -35,6 +35,10 @@ jobs:
           run: npm i
 
 #Generation
+        - name: Discover ts declarations
+          working-direcory: devextreme
+          run: npm run discover-ts-declarations -- --artifacts ../artifacts --js-scripts ../devextreme/js
+
```
IlyaKhD: It seems that this lines are redundant:
```suggestion
        - name: Discover ts declarations
          working-direcory: devextreme
          run: npm run discover-ts-declarations -- --artifacts ../artifacts --js-scripts ../devextreme/js

```

next step calls `npm run discover-declarations` script from devextreme, which is: 
```
"discover-declarations": "dx-tools discover --js-scripts ./js --exclude js/renovation/",
```
which calls tsDiscoverer in it's turn:
```
'discover': [runTsDiscoverer, runDiscoverer],
```

Igggr: Yes. Let's rename this script to discover - we have `discover-declarations` in devextreme-internal-tools -> `index.ts` that call Declarations.dll. So such naming is confusing

IlyaKhD:
Closed due to #3868

PR #3491, opened 2022-03-03
Change the activeStateEnabled description

vladaskorohodova:


api-reference/10 UI Components/Widget/1 Configuration/activeStateEnabled.md
```diff
@@ -5,7 +5,9 @@ default: false
 ---
 ---
 ##### shortDescription
-Specifies whether or not the UI component changes its state when interacting with a user.
+Specifies whether the UI component changes its state as a result of user interaction.
 
 ---
-This property is used when the UI component is displayed on a platform whose guidelines include the active state change for UI components.
+The UI component switches to active state when users press down the primary mouse button. When this property is set to **true**, the CSS rules for the active state apply. You can change these rules to customize the component.
```
albertov05: `The UI component switches to active state when users press down the primary mouse button.`

1. Please add an article ("the") before "active state": "the active state".

2. Does "press down" mean that the user presses the mouse button down, but does not release it? Just checking for my understanding.

api-reference/10 UI Components/Widget/1 Configuration/activeStateEnabled.md
```diff
@@ -5,7 +5,9 @@ default: false
...
+
+Use this property when you display the component on a platform whose guidelines include the active state change for UI components. See the example of such platform: [MUI](https://github.com/mui/material-ui).
```
albertov05:
```suggestion
Use this property when you display the component on a platform whose guidelines include the active state change for UI components. See the following topic for an example of this type of platform: [MUI](https://github.com/mui/material-ui).
```

api-reference/10 UI Components/Widget/1 Configuration/activeStateEnabled.md
vladaskorohodova: Yes, it does mean exactly that. Should I add something like "when users press down the primary mouse button but do not release it"?

api-reference/10 UI Components/Widget/1 Configuration/activeStateEnabled.md
vladaskorohodova: I think that "topic" is not the right word, because the link goes to the official github repository of the platform, not the article. Can I write "See the following link" or something like that?

api-reference/10 UI Components/Widget/1 Configuration/activeStateEnabled.md
albertov05: Thanks for the clarification. I just wanted to know for my own understanding. We can leave your version as is. :)

api-reference/10 UI Components/Widget/1 Configuration/activeStateEnabled.md
albertov05: Thanks for the clarification. How about the following?

`Use this property when you display the component on a platform whose guidelines include the active state change for UI components. See the following GitHub repository for an example of this type of platform: [MUI](https://github.com/mui/material-ui).`

artem-kurchenko (approved):

PR #3649, opened 2022-04-25
Notify: add description

vladaskorohodova:


api-reference/50 Common/utils/ui/notify(message_stack).md
```diff
@@ -0,0 +1,136 @@
...
+---
+#include common-demobutton with {
+    url: "https://js.devexpress.com/Demos//WidgetsGallery/Demo/Toast/Stack/"
+}
+
+To stack toast messages, specify the `position` field in the **stack** object.
+
+The `position` field can be one of the following:
+
+- A string
```
AnnaKondratova:
```suggestion
- String    
```

api-reference/50 Common/utils/ui/notify(message_stack).md
```diff
@@ -0,0 +1,136 @@
...
+
+    - *'right center'*
+
+    - *'bottom left'*
+
+    - *'bottom center'*
+
+    - *'bottom right'*
+
+- An object
```
AnnaKondratova:
```suggestion
- Object    
```

AnnaKondratova (approved):

api-reference/50 Common/utils/ui/notify(message_stack).md
```diff
@@ -0,0 +1,136 @@
...
+    - *'bottom center'*
+
+    - *'bottom right'*
+
+- Object    
+Avaliable values: *{left | right, top | bottom}*.
+
+[note]
+
+If you use an object for `position` field, you can only specify one field in that object for vertical coordinate and one field for horizontal coordinate (for example, if you specify *'top'*, you do not need to specify *'bottom'*).
```
Abadzhev: This explanation sounds a bit strange. I understand that you want to comment on the syntax you used above. But I don't think I would ever try to set a value that reads "right left top". That doesn't make much sense :-) 

On the other hand, I see that "center" values are missing if you specify position via an object. That's maybe something that's worth an emphasis. 

What if you use a single table to list possible values? Then you display all options in one place. You don't need the note. And you also emphasize that String values give you a few extra options. 

| String | Object | 
| ------------- | ------------- |
| `'top left'` | `{left, top}` |
| `'top center'` | N/A |
| `'top right'` | `{right, top}` |
| `'left center'` | N/A |
| `'center'` | N/A |
| `'right center'` | N/A |
| `'bottom left'` | `{left, bottom}` |
| `'bottom center'` | N/A |
| `'bottom right'` | `{right, bottom}` |

This comment applies to the other topic as well.

api-reference/50 Common/utils/ui/notify(message_stack).md
```diff
@@ -0,0 +1,136 @@
...
+
+- *'up'*
+
+- *'down'*
+
+- *'left'*
+
+- *'right'*
+
+- *'up-reverse'*
```
Abadzhev: Can we still change these names? Later in the topic, you explain the difference well. You use the words "push" and "stack" that make much more sense to me than "up" and "up-reverse" (that just sounds like a bad name for "down" :-) ). So it would be great if we can change these options to `up-push`, `up-stack`, `down-push`, `down-stack`, and so on.

(This comment applies to the other topic as well.)

api-reference/50 Common/utils/ui/notify(options_stack).md
```diff
@@ -0,0 +1,138 @@
...
+
+The image below shows the difference between directions with and without the 'reverse' postfix:
+
+![Toast Messages When Direction Is 'Up'](/images/Utilities/notify-stack-up.png) ![Toast Messages When Direction Is 'Up-Reverse'](/images/Utilities/notify-stack-up-reverse.png)
+
+When you set the `direction` to *'up'*, the second and subsequent toasts push the first toast upwards. When you set the `direction` to *'up-reverse'*, toasts stack on top of each other.
+
+If you do not specify the `direction` field, its default value is *'down'* for `position: 'top left | top center | top right'`, and *'up'* for other `position` options.
+
+Note that the default [hide](/Documentation/ApiReference/UI_Components/dxToast/Configuration/animation/#hide) animation duration for toast messages is 400 milliseconds. When you specify the `direction` with the 'reverse' postfix, set the **hide** animation duration to a lesser number or disable it to prevent the last toast in a stack from staying longer on the screen.
```
Abadzhev: Why didn't you put the same note in the other topic? I thought these were the same toast messages with the same settings. 

I have two comments for this text: 

- This sounds like a bug to me, especially when written like this. If I choose to stack notifications, then I assume that when the first one disappears, the rest shift closer to the edge. They don't remain in their positions. In other words, I would get stuck with a toast that was fourth in line and now hangs there alone in the middle of my screen. So... I don't mind that you want to mention the option, but I would use more neutral wording in hope that this behavior changes for the better in future releases. (That is, don't mention that there can be a single toast message staying in the middle of the screen for a while.) 
- "set to a lesser number". If you see such lengthy constructions, you can usually find a single verb to replace them. For example, I used "reduce". 

```suggestion
Note that toast messages disappear with an animated effect after their display time ends. The default [hide](/Documentation/ApiReference/UI_Components/dxToast/Configuration/animation/#hide) animation duration for toast messages is 400 milliseconds. If multiple messages can stack in your application and block important visual elements, you may want to disable the animation or reduce its duration.
```

api-reference/50 Common/utils/ui/notify(message_stack).md
vladaskorohodova: I decided not to change these names.
The main reason is I think that the 'reverse' postfix is familiar to our customers. It is used in [CSS](https://developer.mozilla.org/ru/docs/Web/CSS/flex-direction) and in [other UI libraries](https://mui.com/material-ui/react-stack/). The other thing, we already use 'stack' as an option name, and I think it's a bit odd to use it 2 times in the same property.

api-reference/50 Common/utils/ui/notify(message_stack).md
```diff
@@ -0,0 +1,151 @@
...
+        <th>String</th>
+        <th>Object</th>
+    </tr>    
+    <tr>
+        <td><code>'top left'</code></td>
+        <td><code>{left, top}</code></td>
+    </tr>
+    <tr>
+        <td><code>'top center'</code></td>
+        <td>Use window's and toast's width to calculate the center of the window, then use the <i>'top'</i> field.</td>
```
Abadzhev, line 40: Maybe we should say "container" instead of "window". I don't necessarily need to display my toast centered in the window. It may be centered within a specific "div" (any area, I guess).

api-reference/50 Common/utils/ui/notify(message_stack).md
Abadzhev: OK - now I see where it came from. But then I should point out a difference between what they do and what we do. 

They use neutral words like "row" and "reverse" to account for right-to-left interfaces/languages. So if you specify "row", then it arranges elements left-to-right or right-to-left, depending on the current culture. So then you can say "row-reverse" and this options turns things around - you'll see right-to-left arrangement in cultures with left-to-right languages, and vice versa. 

(This is like WinForms has an alignment enumeration with values Near, Center, and Far. Their meaning changes based on whether the app display left-to-right or right-to-left content. Near usually means Left, but it means Right in RTL languages.)

Then, I guess, they decided to do "column" and "column-reverse" to be consistent. They could have named them "down" and "up" (because that's what they always are - regardless of other settings).

To summarize - those CSS options don't mix specific directions (up, down, left, right) with the word "reverse". That would be confusing, because "left-reverse" is "right".

I could see the point in using the word "reverse" if these values were dependent on the culture. Then you would also need to use vague words like "column" and "row" (or "vertical" and "horizontal"). But if these values specify specific directions (as it seems to me), then I would just name them in a way that eliminates confusion. If you don't want to use the word "stack", you can simply go with `up` and `up-push`.

Obviously this is a small control and it's not the end of the world if you name settings one way or another... But these little things can accumulate into overall impression of simplicity or frustration. So if you can consider my arguments - great. If you decide to keep things as is... Well, I tried :-)

api-reference/50 Common/utils/ui/notify(message_stack).md
```diff
@@ -0,0 +1,151 @@
...
+To stack toast messages, specify the `position` field in the **stack** object.
+
+The `position` field can be one of the following:
+
+<table class="dx-table">
+    <tr>
+        <th>String</th>
+        <th>Object</th>
+    </tr>    
+    <tr>
```
Abadzhev, line 30: As I understand, Object allows me to set coordinates as numbers. I didn't realize that. Then maybe you can dedicate a separate table row to that. Here's a suggestion:

```suggestion
    <tr>
        <td>Cannot specify coordinates</td>
        <td><code>{x, y}</code>, e.g. <code>{100, 50}</code></td>
    </tr>    
    <tr>
```

api-reference/50 Common/utils/ui/notify(message_stack).md
vladaskorohodova: The toast is always positioned realtively to the window (as the developer said), so that's why I'm writing 'window'.

api-reference/50 Common/utils/ui/notify(message_stack).md
vladaskorohodova: I don't like `up` and `up-push` for the reason that it does not convey what `up` means. I'm okay with `up-push`, but I think we need another word for `up-reverse`.
I tried to come up with suggestions, please check them out:
`up-load` (I know it sounds like upload, but still...)
`up-layer`
`up-compile`
`up-extend`

Abadzhev: @vladaskorohodova 

I was thinking about  `up-add` or 'up-append' as an opposite to `up-push`. 

I still think the best choice would be `up-stack`. "Stack" is what kids do when they put blocks on top of each other. I think that word describes the process best. You didn't want to have "stack" in both the option name and the option value though. 

In your versions, I was thinking about `up-layer`, but I would always think of 'layer' as a noun first. `up-extend` is probably the best choice from your list. 

So I would go with `up-add`, `up-append`, or `up-extend`. My personal favorite is `up-append`, but I can live with any of the three.

vladaskorohodova: We had a chat with Anna Kondratova and decided to use `up-push` and `up-stack`. I guess I can live with 'stack' being used 2 times...

api-reference/50 Common/utils/ui/notify(message_stack).md
```diff
@@ -0,0 +1,155 @@
...
+id: ui.notify(message,stack)
+module: ui/notify
+export: default
+---
+---
+##### shortDescription
+Creates a stackable toast message.
+
+##### param(message): String
+The message's text.
```
dirk-pieterse:
```suggestion
The message content.
```

dirk-pieterse (approved):

PR #3687, opened 2022-05-05
NumberBox: add description to the 'format' property

vladaskorohodova:


api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+
+- Predefined format    
+
+        <!-- tab: JavaScript -->
+        const number = 1.2345;
+
+        type: "fixedPoint" // 1
+        type: "decimal" // 1.2345
+        type: "percent" // 123%
+        type: "currency" // $1
```
Abadzhev, line 25: All currency formats usually have 2 digits after the decimal point. The default formatting for "currency" should be $1.23 (this is probably more of a comment to relay to developers).

You can try it in Excel - the default value for Decimal Places will be 2.

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+
+    You can specify the [precision](/Documentation/ApiReference/Common/Object_Structures/format/#precision) property for decimal numbers.
+
+    NumberBox does not support large number formats. If you want to apply these formats, you can implement the [formatter](/api-reference/50%20Common/Object%20Structures/format/formatter.md '/Documentation/ApiReference/Common/Object_Structures/format/#formatter') and [parser](/api-reference/50%20Common/Object%20Structures/format/parser.md '/Documentation/ApiReference/Common/Object_Structures/format/#parser') functions or specify a custom format string.
```
Abadzhev: I'm not sure what you mean by "large number format". Does Excel have an analogy? Maybe you mean Scientific: 1.23E+00?

When you mention "custom format string" I would add "as described below" for more clarity.

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+    Use Unicode Locale Data Markup Language (LDML) patterns to specify a custom format string. An LDML pattern consists of wildcard characters and characters displayed as is. The **format** property supports the following wildcard characters:
+
+    <table class="dx-table">
+        <tr>
+        <th>Format character</th>
+        <th>Description</th>
+        </tr>
+        <tr>
+        <td>0</td>
+        <td>A digit. Displays '0' if it is not specified in the UI.</td>
```
Abadzhev:
```suggestion
        <td>A digit. Displays '0' if the formatted number doesn't have a digit in that position.</td>
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+        </tr>
+        <tr>
+        <td>&#35;</td>
+        <td>
+                A digit or nothing. One symbol represents several integer digits, but only one decimal digit.</br>
```
Abadzhev: Adding a few details. 

```suggestion
                Any number of leading digits, a single digit, or nothing. If this character goes first in the format string, it can match multiple leading digits (before the decimal point). Consequent characters match a single digit. If the formatted number doesn't have a digit in the corresponding position, displays nothing.</br> 
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+        </tr>
+        <tr>
+        <td>0</td>
+        <td>A digit. Displays '0' if it is not specified in the UI.</td>
+        </tr>
+        <tr>
+        <td>&#35;</td>
+        <td>
+                A digit or nothing. One symbol represents several integer digits, but only one decimal digit.</br> 
+                For example, "#0.#" represents "123.4", but not "123.45".
```
Abadzhev: I think it would be a bit more straightforward if we specify the source number, the format, and the result. The point is that the leading character can match multiple digits and this sentence will still illustrate that concept. 

Another note - try to avoid the word "represents". We tend to use it too frequently, but not always correctly. 

```suggestion
                For example, if you apply format "#0.#" to "123.45", the result is "123.4".
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+        </td>
+        </tr>
+        <tr>
+        <td>.</td>
+        <td>
+                A decimal separator. </br>
+                Displayed according to the specified locale.
```
Abadzhev: Got rid of passive and ING. 

```suggestion
                Actual character depends on locale.
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+        </td>
+        </tr>
+        <tr>
+        <td>,</td>
+        <td>
+                A group separator. </br>
+                Displayed according to the specified locale.
```
Abadzhev: Same comment. 

```suggestion
                Actual character depends on locale.
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+        </td>
+        </tr>
+        <tr>
+        <td>%</td>
+        <td>
+                The percent sign. Multiplies the input value by 100. </br>    
+                If it is enclosed in single quotes ('%'), it only adds this sign to the input value.
```
Abadzhev: I think you should add another row to the table where you will describe quote characters. Define them as "escape characters" that help you display any wildcards as literals. 

If you do that, then you can remove this note from this table row. Really, this applies not only to the percent character.

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+        </td>
+        </tr>
+        <tr>
+        <td>;</td>
+        <td>Separates positive and negative numbers. If there is no explicit negative format, a positive number receives the "-" prefix. </td>
```
Abadzhev: Let's be a bit more clear about what we specify (not positive and negative numbers, but formats).

As a rule of thumb, try to avoid "there is", "there are", "it is", etc. Better try and find a more specific subject-verb pair. 

```suggestion
        <td>Separates positive and negative format patterns. This character is optional. If you don't use it, then you specify a common format for positive and negative numbers. Negative numbers simply display a minus ("-") prefix. </td>
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+        </tr>
+        <tr>
+        <td>Other characters</td>
+        <td>
+                Any character. Should be placed only at the format string's beginning or end. </br>
+                You can use the special characters above as well (in single quotation marks).
```
Abadzhev: If you add a separate row about "escape characters", then you wouldn't need this second line.

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+        </td>
+        </tr>
+        <tr>
+        <td>;</td>
+        <td>Separates positive and negative numbers. If there is no explicit negative format, a positive number receives the "-" prefix. </td>
+        </tr>
+        <tr>
+        <td>Other characters</td>
+        <td>
+                Any character. Should be placed only at the format string's beginning or end. </br>
```
Abadzhev: I avoid the passive and also use the term "literal character" to indicate that those characters appear "as is". 
 
```suggestion
                You can add any literal characters to the beginning or end of the format string. </br>
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+        <!-- tab: JavaScript -->
+        const largeNumber = 123456.789;
+
+        // Limit the decimal part to one digit
+        format: "#0.#" // 123456.7
+
+        // Add a group separator
+        format: ",##0.###" // 123,456.789
+        
+    The example below shows how to specify percentages when you work with decimals. It is necessary to add a `#` character at the start of an LDML pattern to avoid digits before `.` to be deleted.
```
Abadzhev: Again - try to avoid sentences that start with "it is". 

But in this case, I would rather remove the second sentence. It simply explains again what we already said in the table. If you want to preserve all your leading digits, then use the hash mark ('#'). 

```suggestion
    The examples below show different ways to apply percentage formatting to decimals. 
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+        // Represent as a percentage and limit to two decimal digits
+        format: "#0.##%" // 1.23%
+
+        // Add a percent sign and limit to two decimal digits
+        format: "#0.##'%'" // 0.01%
+
+        const number = 0.80785;
+
+        // Remove all the numbers except the last digit before '.'
+        format: "0.##%" // 0.79%, would be 80.79% if the format is "#0.##%"
```
Abadzhev: I don't like this example - you probably would never want to format numbers like that. I suggest you remove that one. 

On the other hand, I understand that you're trying to make a point. You want to illustrate the risks of using '0' as the first character. If you really want to keep this example, then I would change its preceding comment to something like this: "Use caution if your format strings starts with a zero ('0'), because the formatted number may lose leading digits".

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,115 @@ default: ''
...
+    
+- Function    
+A function should apply a custom format to a value and return the formatted value as a string. Functions are useful for advanced formatting. Refer to the following article for more information: [Custom Function](/Documentation/Guide/Common/Value_Formatting/#Format_UI_Component_Values/Custom_Function).
+
+- Object    
+Full format configuration. The object structure is shown in the [format](/Documentation/ApiReference/Common/Object_Structures/format/) API section.
+
+[note] With this property specified, a press on Minus Sign (-) inverts the current value instead of entering "-".
+
+[note] If you set this property, the telephone keyboard is used for editing on mobile devices. However, it may not have a point, comma, or other symbols for entering decimals. Set the [mode](/api-reference/10%20UI%20Components/dxNumberBox/1%20Configuration/mode.md '/Documentation/ApiReference/UI_Components/dxNumberBox/Configuration/#mode') property to *"text"* to use the standard keyboard instead.
```
Abadzhev: I'm not sure that it's always a "telephone" thing. I think a "number pad" is a better name. 

In the second sentence you don't need to list all versions of decimal characters. The word "however" also seems redundant to me. 

```suggestion
[note] If you set this property, the number pad keyboard appears on mobile devices when users focus the editor. Such keyboards may not have a decimal character. Set the [mode](/api-reference/10%20UI%20Components/dxNumberBox/1%20Configuration/mode.md '/Documentation/ApiReference/UI_Components/dxNumberBox/Configuration/#mode') property to *"text"* to use the standard keyboard instead.
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
vladaskorohodova: The developers do not have plans for changing the default 'currency' format. 
Users can set `format: { type: 'currency', precision: 2 }` if they want to have 2 digits after the decimal point. I mention the 'precision' property in the sentence below.
Should I add an example here so it becomes more clear?

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
vladaskorohodova: There is a global type [format](https://js.devexpress.com/Documentation/ApiReference/Common/Object_Structures/format/#type), which supports small and large numbers. I really didn't know how to adequately mention the fact that NumberBox does not support large number format types without showing the types in this description. I think I don't need to show them here, since customers can't use them anyway. If I add a link, e.g. [large number formats](https://js.devexpress.com/Documentation/ApiReference/Common/Object_Structures/format/#type), would it be enough?

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
vladaskorohodova: Noted

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
vladaskorohodova: I decided to leave the caution message before the code snippet and delete the last example.

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
Abadzhev: One way to deal with it, is simply not mention those formats. I don't expect a lot of people to care about those formats. We draw attention to the fact that something may not be supported. 

Or maybe phrase is in a bit more positive way - turn the order around. (Don't say "something is not supported". Everything is supported, you just need to know how to enable it :-) ) And then you better add a link, yes. 

```suggestion
    If you want to apply a different format, you can implement the [formatter](/api-reference/50%20Common/Object%20Structures/format/formatter.md '/Documentation/ApiReference/Common/Object_Structures/format/#formatter') and [parser](/api-reference/50%20Common/Object%20Structures/format/parser.md '/Documentation/ApiReference/Common/Object_Structures/format/#parser') functions or specify a custom format string. For example, you can use this functionality to enable [large number formats](/Documentation/ApiReference/Common/Object_Structures/format/#type).
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
Abadzhev: That's weird... Oh well, we tried :-) It's out of our hands then.

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,114 @@ default: ''
...
+        const number = 1.2345;
+
+        type: "fixedPoint" // 1
+        type: "decimal" // 1.2345
+        type: "percent" // 123%
+        type: "currency" // $1
+
+    You can specify the [precision](/Documentation/ApiReference/Common/Object_Structures/format/#precision) property for decimal numbers.
+
+    If you want to apply a different format, you can implement the [formatter](/api-reference/50%20Common/Object%20Structures/format/formatter.md '/Documentation/ApiReference/Common/Object_Structures/format/#formatter') and [parser](/api-reference/50%20Common/Object%20Structures/format/parser.md '/Documentation/ApiReference/Common/Object_Structures/format/#parser') functions or specify a custom format string. For example, you can use this functionality to enable [large number formats](/Documentation/ApiReference/Common/Object_Structures/format/#type).
```
vladaskorohodova:
```suggestion
    If you want to apply a different format, you can implement the [formatter](/api-reference/50%20Common/Object%20Structures/format/formatter.md '/Documentation/ApiReference/Common/Object_Structures/format/#formatter') and [parser](/api-reference/50%20Common/Object%20Structures/format/parser.md '/Documentation/ApiReference/Common/Object_Structures/format/#parser') functions or specify a custom format string as described below. For example, you can use this functionality to enable [large number formats](/Documentation/ApiReference/Common/Object_Structures/format/#type).
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,114 @@ default: ''
...
+        <th>Description</th>
+        </tr>
+        <tr>
+        <td>0</td>
+        <td>A digit. Displays '0' if the formatted number doesn't have a digit in that position.</td>
+        </tr>
+        <tr>
+        <td>&#35;</td>
+        <td>
+                Any number of leading digits, a single digit, or nothing. If this character goes first in the format string, it can match multiple leading digits (before the decimal point). Consequent characters match a single digit. If the formatted number doesn't have a digit in the corresponding position, displays nothing.</br>
```
albertov05:
```suggestion
                Any number of leading digits, a single digit, or nothing. If this character goes first in the format string, it can match multiple leading digits (before the decimal point). Subsequent characters match a single digit. If the formatted number doesn't have a digit in the corresponding position, it displays nothing.</br>
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,114 @@ default: ''
...
+        </tr>
+        <tr>
+        <td>%</td>
+        <td>
+                The percent sign. Multiplies the input value by 100.
+        </td>
+        </tr>
+        <tr>
+        <td>;</td>
+        <td>Separates positive and negative format patterns. This character is optional. If you do not use it, then you specify a common format for positive and negative numbers. Negative numbers simply display a minus ("-") prefix. </td>
```
albertov05: `Negative numbers simply display a minus ("-") prefix.`

Is this sentence an example of a common format for negative numbers? In other words, do we need to add "For example" to the beginning of this sentence? Please clarify.

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,114 @@ default: ''
...
+        <!-- tab: JavaScript -->
+        const largeNumber = 123456.789;
+
+        // Limit the decimal part to one digit
+        format: "#0.#" // 123456.7
+
+        // Add a group separator
+        format: ",##0.###" // 123,456.789
+        
+    The examples below show different ways to apply percentage formatting to decimals. Use caution if your format strings starts with a zero ('0'), because the formatted number may lose leading digits.
```
albertov05:
```suggestion
    The examples below show different ways to apply percentage formatting to decimals. Use caution if your format string starts with a zero ('0'), because the formatted number may lose leading digits.
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,114 @@ default: ''
...
+        format: "#0.##'%'" // 0.01%
+
+    
+- Function    
+A function should apply a custom format to a value and return the formatted value as a string. Functions are useful for advanced formatting. Refer to the following article for more information: [Custom Function](/Documentation/Guide/Common/Value_Formatting/#Format_UI_Component_Values/Custom_Function).
+
+- Object    
+Full format configuration. The object structure is shown in the [format](/Documentation/ApiReference/Common/Object_Structures/format/) API section.
+
+[note] With this property specified, a press on Minus Sign (-) inverts the current value instead of entering "-".
```
albertov05:
```suggestion
[note] With this property specified, a press on the minus sign (-) inverts the current value instead of entering "-".
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
vladaskorohodova: I'm going to rewrite this section and add an example, so it becomes more clear

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,117 @@ default: ''
...
+        <tr>
+        <td>%</td>
+        <td>
+                The percent sign. Multiplies the input value by 100.
+        </td>
+        </tr>
+        <tr>
+        <td>;</td>
+        <td>    Separates positive and negative format patterns.  </br>
+                For example, the "#0.##;(#0.##)" format displays positive number according to the pattern before ";", and negative number according to the pattern after the ";". </br>
```
albertov05:
```suggestion
                For example, the "#0.##;(#0.##)" format displays a positive number according to the pattern before the semicolon (";"), and a negative number according to the pattern after the semicolon (";").</br>
```

api-reference/10 UI Components/dxNumberBox/1 Configuration/format.md
```diff
@@ -8,13 +8,117 @@ default: ''
...
+
+        <!-- tab: JavaScript -->
+        const number = 1.2345;
+
+        type: "fixedPoint" // 1
+        type: "decimal" // 1.2345
+        type: "percent" // 123%
+        type: "currency" // $1
+
+    You can specify the [precision](/Documentation/ApiReference/Common/Object_Structures/format/#precision) property for decimal numbers.
```
vladaskorohodova:
```suggestion
    You can specify the [precision](/Documentation/ApiReference/Common/Object_Structures/format/#precision) property for fractional numbers.
```

DokaRus (approved):
LGTM

PR #3738, opened 2022-05-23
Format: describe useCurrencyAccountingStyle

vladaskorohodova:


api-reference/50 Common/Object Structures/format/useCurrencyAccountingStyle.md
```diff
@@ -5,7 +5,12 @@ default: true
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies whether to apply the accounting style to formatted numbers of the `currency` [type](/Documentation/ApiReference/Common/Object_Structures/format/#type).
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+The accounting style adds parentheses to negative numbers instead of the minus sign. To disable the accounting style in the component and display formatted numbers of the `currency` [type](/Documentation/ApiReference/Common/Object_Structures/format/#type) with the minus sign, set this property to **false**. This property overrides the [global config](/Documentation/ApiReference/Common/Object_Structures/globalConfig/#defaultUseCurrencyAccountingStyle).
```
dmitry-eliseev-devexpress: 1. Do I get this right that if I leave this property unspecified, the resulting style will be determined by the global accounting style flag? If this is correct, I believe we should explain it in a more explicit manner.
2. The global **config**. "Config" does not sound as an ordinary noun, this is a shorthand of some sort from the sound of it. Is it a global configuration file? Or just a set of settings? Is "global config" a proper noun? If you call it "global config" everywhere across your documentation and customers understand it, then I have no objections. I just had to clarify this.

api-reference/50 Common/Object Structures/globalConfig/defaultUseCurrencyAccountingStyle.md
```diff
@@ -5,7 +5,18 @@ default: true
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies whether to apply the accounting style to formatted numbers of the `currency` [type](/Documentation/ApiReference/Common/Object_Structures/format/#type).
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+The accounting style adds parentheses to negative numbers instead of the minus sign. To disable the accounting style globally and display formatted numbers of the `currency` [type](/Documentation/ApiReference/Common/Object_Structures/format/#type) with the minus sign, set this property to **false**.
```
dmitry-eliseev-devexpress, line 11: This topic does not include a connection to the corresponding "local" accounting style flag that corresponds to individual components and has higher priority than the currently described setting. I don't believe that a See Also link is enough to cover that.

dmitry-eliseev-devexpress (commented):
We need to make some things clear before moving on.

api-reference/50 Common/Object Structures/format/useCurrencyAccountingStyle.md
vladaskorohodova: I tried to explain it better and also added an image in both files to illustrate the link between these properties.

Also I removed the word "config".

api-reference/50 Common/Object Structures/globalConfig/defaultUseCurrencyAccountingStyle.md
```diff
@@ -5,7 +5,19 @@ default: true
...
+
+    DevExpress.config({ defaultUseCurrencyAccountingStyle: false });
+    // ===== or when using modules =====
+    import config from "devextreme/core/config";
+        
+    config({
+        defaultUseCurrencyAccountingStyle: false 
+    });
+
+This property specifies the global configuration. If you want to override this setting in the component, specify the [useCurrencyAccountingStyle](/Documentation/ApiReference/Common/Object_Structures/format/#useCurrencyAccountingStyle) property in the [format](/Documentation/ApiReference/Common/Object_Structures/format/) object.
```
dmitry-eliseev-devexpress: In **any** component, am I right? As for the `format` object, to which component it belongs? Does every component has its own `format` object? Right now it's not quite clear.

```suggestion
This property specifies the global configuration. If you want to override this setting in a component, specify the [useCurrencyAccountingStyle](/Documentation/ApiReference/Common/Object_Structures/format/#useCurrencyAccountingStyle) property in the [format](/Documentation/ApiReference/Common/Object_Structures/format/) object.
```

api-reference/50 Common/Object Structures/format/useCurrencyAccountingStyle.md
```diff
@@ -5,7 +5,16 @@ default: true
...
-<!-- Description goes here -->
\ No newline at end of file
+The accounting style adds parentheses to negative numbers instead of the minus sign. To disable the accounting style in the component and display formatted numbers of the `currency` [type](/Documentation/ApiReference/Common/Object_Structures/format/#type) with the minus sign, set this property to **false**.
+
+    format({
+        type: "currency",
+        useCurrencyAccountingStyle: false 
+    });
+
+If you do not specify this property, the [defaultUseCurrencyAccountingStyle](/Documentation/ApiReference/Common/Object_Structures/globalConfig/#defaultUseCurrencyAccountingStyle) global configuration setting is applied. This property overrides the global configuration.
```
dmitry-eliseev-devexpress:
```suggestion
If you do not specify this property, the [defaultUseCurrencyAccountingStyle](/Documentation/ApiReference/Common/Object_Structures/globalConfig/#defaultUseCurrencyAccountingStyle) global configuration setting is in effect. The `useCurrencyAccountingStyle` property has priority over the global setting.
```

dmitry-eliseev-devexpress (approved):
Almost good to go.

api-reference/50 Common/Object Structures/globalConfig/defaultUseCurrencyAccountingStyle.md
vladaskorohodova: Some of the components have the `format` property, which has a common (for all the components) data type also called `format`. The `format` can be declared as a string, object, or a function. Basically I tried to tell users something like this: if your component value can be formatted, i.e. the component has the `format` property in its configuration, you can specify the useCurrencyAccountingStyle option using specifically an object as a declaration for the `format` property.

dmitry-eliseev-devexpress: OK, I see now. That means it should be something like that:

```suggestion
This property specifies the global accounting style setting. If you want to override this setting for a component, specify the [useCurrencyAccountingStyle](/Documentation/ApiReference/Common/Object_Structures/format/#useCurrencyAccountingStyle) property in the component's [format](/Documentation/ApiReference/Common/Object_Structures/format/) object.
```

dmitry-eliseev-devexpress (approved):
Now we're good to go. Thank you.

dirk-pieterse (approved):

PR #3777, opened 2022-06-02
Scheduler: Get CellData of multiple selected cells after user actions T956636

vladaskorohodova:


api-reference/10 UI Components/dxScheduler/1 Configuration/selectedCellData.md
```diff
@@ -21,6 +21,15 @@ This array contains objects with the following structure:
         }     
     }
```
albertov05, line 23: I think we should update line 8 above to avoid the plural possesive noun. Can we update it as follows?

Line 8:
The data of the currently selected cells.

vladaskorohodova: Noted

AnnaKondratova (approved):

PR #3940, opened 2022-07-01
Grids: update the 'setCellValue' description

vladaskorohodova:


api-reference/_hidden/GridBaseColumn/setCellValue.md
```diff
@@ -19,7 +19,7 @@ A read-only parameter providing access to the current row data.
 Return a promise for an asynchronous operation or return nothing.
```
albertov05, line 20: Line 7 (above):
`Specifies a function to be invoked after the user has edited a cell value, but before it will be saved in the data source.`

Can we update this sentence as follows?

`Specifies a function to be invoked after the user has edited a cell value, but before it is saved in the data source.`

api-reference/_hidden/GridBaseColumn/setCellValue.md
```diff
@@ -170,7 +170,9 @@ This function allows you to process user input before it is saved to the data so
 
 ---
 
-If you need to perform asynchronous operations in the **setCellValue** function, return a promise from it. In the following code, this technique is used to get the `Tax` value from the server when the `State` value is changed:
+If you use **setCellValue** and the [editing]({basewidgetpath}/Configuration/editing/).[mode]({basewidgetpath}/Configuration/editing/#mode) is `form` or `popup`, the **setCellValue** forces the component to repaint all items in the editing form. Set the [repaintChangesOnly]({basewidgetpath}/Configuration/#repaintChangesOnly) to **true** so the component re-renders only the fields with the changed values.
```
albertov05:
```suggestion
If you use **setCellValue** and the [editing]({basewidgetpath}/Configuration/editing/).[mode]({basewidgetpath}/Configuration/editing/#mode) is `form` or `popup`, the **setCellValue** forces the component to repaint all items in the edit form. Set [repaintChangesOnly]({basewidgetpath}/Configuration/#repaintChangesOnly) to **true** so the component re-renders only the fields with the changed values.
```

api-reference/_hidden/GridBaseColumn/setCellValue.md
vladaskorohodova: Okay

PR #4135, opened 2022-08-26
Sliders: describe callValueChange option

vladaskorohodova:


api-reference/10 UI Components/dxSliderBase/1 Configuration/callValueChange.md
```diff
@@ -6,7 +6,139 @@ default: 'onMoving'
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies when to call the [onValueChanged]({basewidgetpath}/Configuration/#onValueChanged) function.
```
ksercs: I think it's better to write something like "specified when to change value". `onValueChanged` handler call is only a side effect of a value change.

api-reference/10 UI Components/dxSliderBase/1 Configuration/callValueChange.md
```diff
@@ -6,7 +6,139 @@ default: 'onMoving'
...
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+If you want to call the [onValueChanged]({basewidgetpath}/Configuration/#onValueChanged) after a user finishes moving the component's handle, set this property to `onMovingComplete`:
```
ksercs: the same here

PR #4196, opened 2022-09-15
CheckBox: describe allowIndeterminateStateByClick

vladaskorohodova:


api-reference/10 UI Components/dxCheckBox/1 Configuration/allowIndeterminateStateByClick.md
```diff
@@ -5,7 +5,19 @@ default: false
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies whether users can set the CheckBox state to indeterminate.
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+The component can be in three states (checked, unchecked, and indeterminate). When the CheckBox [value](/Documentation/ApiReference/UI_Components/dxCheckBox/Configuration/#value) is `null` or `undefined`, the default state is indeterminate.
```
Abadzhev: "can be in"... The verb "to be" is considered a weak verb. I would replace it with something like "supports".

The you talk about the "default" state. I don't think it's 100% accurate. I guess I can set `value` to `null` in code for a number of reasons. So it's not only a possible "default" state, it's just a possible state. I understand that use this to introduce an example later, but I'll deal with that too. 

```suggestion
*dxCheckBox* supports three states: checked, unchecked, and indeterminate. The following table specifies the relation between the editor's state and its [value](/Documentation/ApiReference/UI_Components/dxCheckBox/Configuration/#value).

| Value | State |
| --- | --- |
| `null` or `undefined` | Indeterminate |
| `true` | Checked |
| `false` | Unchecked |

File: api-reference/10 UI Components/dxCheckBox/1 Configuration/allowIndeterminateStateByClick.md [hunk 2]
```diff
@@ -5,7 +5,19 @@ default: false
...
+
+If this property is set to `false`, users can change checkbox states in the following order:
```
Abadzhev: I want to mention the property name. If you say "this property", then someone may think you're talking about  the `value` property that you mentioned in the previous paragraph. 

I think it's easier to start with the simplest example - all states are available. 
 
```suggestion
If you set *allowIndeterminateStateByClick* to `true`, users can cycles through check box states in the following order:
```

api-reference/10 UI Components/dxCheckBox/1 Configuration/allowIndeterminateStateByClick.md
```diff
@@ -5,7 +5,19 @@ default: false
...
+
+*Indeterminate (default) → Checked → Unchecked → Checked → Unchecked.*
```
Abadzhev:
```suggestion
*Indeterminate → Checked → Unchecked → Indeterminate →  ...*
```

api-reference/10 UI Components/dxCheckBox/1 Configuration/allowIndeterminateStateByClick.md
```diff
@@ -5,7 +5,19 @@ default: false
...
+
+Set this property to `true` to allow users to set the indeterminate state:
```
Abadzhev: Now we have arrived at a more complicated example. 

```suggestion
If you set *allowIndeterminateStateByClick* property to `false`, users can only cycle between the checked and unchecked states. Note that a check box can initially appear in the indeterminate state, if you haven't set the value to either `true` or `false`.
```

api-reference/10 UI Components/dxCheckBox/1 Configuration/allowIndeterminateStateByClick.md
```diff
@@ -5,7 +5,19 @@ default: false
...
+
+*Indeterminate (default) → Checked → Unchecked → Indeterminate.*
```
Abadzhev:
```suggestion
*Indeterminate → Checked → Unchecked → Checked → Unchecked → ...*
```

api-reference/10 UI Components/dxCheckBox/1 Configuration/value.md
```diff
@@ -11,13 +11,13 @@ Specifies the UI component state.
...
 - unchecked  
- **value** is **false**
+ **value** is `false`
 
-- undetermined  
- **value** is **undefined** or **null**
+- indeterminate  
+ **value** is `undefined` or `null`
 
 Users can press the **Space** key to change the **value**.
```
Abadzhev, line 22: This note makes me wonder if the following option name is good `allowIndeterminateStateByClick`. Maybe a better name would be `allowIndeterminateState`. Even "click" is not always a click - with all the mobile devices, it may be a tap.

vladaskorohodova: Me and developers had a long discussion about the name and we decided to use this one because when you simply say "allowIndeterminateState" - it's not true. It's already allowed. Like you already can set **value** to `null` or `undefined` and the CheckBox will be in the indeterminate state. It may confuse customers. They may think that we don't allow indeterminate state at all, and you need to enable the option to use the state, which is not true.
I think that clicking on a check box is the main way users interact with the component. Like yes, there are mobile devices, but when we use event handlers - we use onClick handler, not onClick and onTap. I also agree that yes, we can use the Space key to change the value, but it's also an additional way to change it, the main is still clicking.

vladaskorohodova: Also maybe as an alternative - allowSwitchToIndeterminateState?

Abadzhev: `allowSwitchToIndeterminateState` - this version is not that much different from `allowIndeterminateState`. We never mention the user side :-)

I know my suggestion wasn't specific enough. I just know that some platforms/controls get away with it. In WinForms, the property is simply called "ThreeState". (Behaves the same way.) 

I can't find a very good name that would imply user interaction, while being concise... Here are a couple of suggestions you can review:

- enableThreeStateBehavior
- enableThreeStateCycle 

But if you'll want to stick with the original name, at least maybe you can go with "...OnClick", instead of "...ByClick".

PR #4275, opened 2022-10-13
DataGrid: Add extra information about using the selectRows() method T…

sergepilipchuk:
…1116836

### DIFF
--- a/api-reference/10 UI Components/GridBase/3 Methods/selectRows(keys_preserve).md
+++ b/api-reference/10 UI Components/GridBase/3 Methods/selectRows(keys_preserve).md
@@ -20,9 +20,9 @@ By default, this method call clears selection of previously selected rows. To ke
 
     <!--JavaScript-->widgetInstance.selectRows([5, 10, 12], true);
 
-If you specify [keyExpr](/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/#keyExpr) as an array of keys, you can call this method as follows:
+If you specify DataGrid's key as composite, for example `key = ['id', 'name']`, you need call this method as follows:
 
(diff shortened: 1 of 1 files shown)


api-reference/10 UI Components/GridBase/3 Methods/selectRows(keys_preserve).md
```diff
@@ -20,6 +20,10 @@ By default, this method call clears selection of previously selected rows. To ke
 
     <!--JavaScript-->widgetInstance.selectRows([5, 10, 12], true);
 
+If you specify DataGrid's key as composite, for example `key = ['id', 'name']`, you need call this method as follows:
```
RomanTsukanov: "For example" should be followed by a comma, but there would be too many commas in this sentence, so I suggest putting the example in brackets.

```suggestion
If you specify DataGrid's key as composite (for example, `key: ['id', 'name']`), you need to call this method as follows:
```

api-reference/10 UI Components/GridBase/3 Methods/selectRows(keys_preserve).md
```diff
@@ -20,6 +20,10 @@ By default, this method call clears selection of previously selected rows. To ke
...
+
+    <!--JavaScript-->widgetInstance.selectRows([{id: 5, name: 'Alex' }, {id: 10: name: 'Bob' }], true);
```
RomanTsukanov:
```suggestion
    <!--JavaScript-->widgetInstance.selectRows([ { id: 5, name: 'Alex' }, { id: 10: name: 'Bob' } ], true);
```

RomanTsukanov (approved):

PR #4486, opened 2022-12-29
Troubleshooting: Bootstrap modals issues

vladaskorohodova:


16adianay (approved):

concepts/80 Troubleshooting/30 3rd-Party Frameworks Issues/00 Bootstrap/05 Modals.md
```diff
@@ -0,0 +1,68 @@
+If you place DevExtreme components in the [Bootstrap modal](https://getbootstrap.com/docs/5.3/components/modal/), the following issues may occur:
```
Abadzhev: I would change the article. We can create many "modals" and place DevExtreme components in any of them. So "in a Bootstrap modal" seems a better choice. This also matches some of the uses from their documentation. 

> Popover in a modal
> Tooltips in a modal


```suggestion
If you place DevExtreme components in a [Bootstrap modal](https://getbootstrap.com/docs/5.3/components/modal/), the following issues may occur:
```

concepts/80 Troubleshooting/30 3rd-Party Frameworks Issues/00 Bootstrap/05 Modals.md
```diff
@@ -0,0 +1,68 @@
...
+
+- Users cannot insert, update, or delete a [DataGrid](/Documentation/Guide/UI_Components/DataGrid/Getting_Started_with_DataGrid/) row.
+- Users cannot enter values in the [FilterBuilder](/Documentation/Guide/UI_Components/FilterBuilder/Overview/).
+- Editors in the [DropDownBox](/Documentation/Guide/UI_Components/DropDownBox/Getting_Started_with_DropDownBox/) lose focus.
+- [TextArea](/Documentation/Guide/UI_Components/TextArea/Getting_Started_with_TextArea/) loses focus.
+- DataGrid [ColumnChooser](/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/columnChooser/) loses focus.
+
+On the page load, Bootstrap places all items inside the modal into a collection and allows to focus only on the items from the collection. DevExtreme component elements can be generated on the fly, so these elements do not get attached to the focusable collection.
```
Abadzhev: "the page load" - article use is a bit harder to explain here. But it's not a pre-determined page load event, it's not "the page load". It may be "a page load"...  but in essence we talk about something generic, so I would use zero article. 

Once again - we shouldn't say "the modal", because I assume I can have a few of those on my page. 

And then I'm breaking longer sentences into shorter sentences and avoiding passive voice. 

```suggestion
On page load, Bootstrap locates all modals and places their children into special collections. A modal only allows items from its children collection to receive focus. At the same time, DevExtreme components can generate elements on the fly. Such dynamically-generated elements do not get attached to the focusable collection.
```

concepts/80 Troubleshooting/30 3rd-Party Frameworks Issues/00 Bootstrap/05 Modals.md
```diff
@@ -0,0 +1,68 @@
...
+
+To prevent this behavior, configure the option that maintains focus inside the modal itself.
```
Abadzhev: When you say "configure the options that maintains focus inside the modal", I assume that you mean "configure" is a synonym to "enable". But... examples set options such as "enforceFocus" to **false**, not to **true**. I would use the word "disable" instead of "configure", just to set the right expectations. 

I also find the phrase "prevent this behavior" a bit vague. Do we prevent the collection construction on page load? Do we prevent the input focus limitation only to collection elements? Or maybe both? I would just use something more specific - the task that the developer has right now. 

```suggestion
To allow any elements to obtain input focus, disable the option that maintains focus inside the modal itself.
```

vladaskorohodova: If you look at the code example for Vue, the option there needs to be set to `true`. So that's why I used "configure", I just cannot choose between 'enable' and 'disable' here. Maybe it's better to use something like:
> To allow any elements to obtain input focus, specify the option that maintains focus inside the modal itself.

Abadzhev: The same issue is back then. If I say "specify the option that shows the filter row", then I automatically think that this means "display the filter row". The way you describe an option ("maintains focus inside"), automatically leads me to believe that this is the action I need to take ("I need to keep the focus inside"). 

So maybe we don't describe the option and only describe the task? Something like this would work for me:

> To allow any element to obtain input focus, set a specially-designed option as shown in code samples below.

concepts/80 Troubleshooting/30 3rd-Party Frameworks Issues/00 Bootstrap/05 Modals.md
```diff
@@ -0,0 +1,68 @@
+If you place DevExtreme components in a [Bootstrap modal](https://getbootstrap.com/docs/5.3/components/modal/), the following issues may occur:
+
+- Users cannot insert, update, or delete a [DataGrid](/Documentation/Guide/UI_Components/DataGrid/Getting_Started_with_DataGrid/) row.
+- Users cannot enter values in the [FilterBuilder](/Documentation/Guide/UI_Components/FilterBuilder/Overview/).
+- Editors in the [DropDownBox](/Documentation/Guide/UI_Components/DropDownBox/Getting_Started_with_DropDownBox/) lose focus.
+- [TextArea](/Documentation/Guide/UI_Components/TextArea/Getting_Started_with_TextArea/) loses focus.
+- DataGrid [ColumnChooser](/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/columnChooser/) loses focus.
+
+On page load, Bootstrap locates all modals and places their children into special collections. A modal only allows items from its children collection to receive focus. At the same time, DevExtreme components can generate elements on the fly. Such dynamically-generated elements do not get attached to the focusable collection.
```
dirk-pieterse: `their children`? Why do you refer to them as 'children'?

dirk-pieterse (approved):

concepts/80 Troubleshooting/30 3rd-Party Frameworks Issues/00 Bootstrap/05 Modals.md
vladaskorohodova: Well, the term "child object" comes from [inheritance](https://en.wikipedia.org/wiki/Inheritance_(object-oriented_programming)#:~:text=through%20inheritance%2C%20a%20%22-,child%20object,-%22%2C%20acquires%20all%20the). In this context, "children" are the inner elements of the modals. That means they become influenced by a modal.
I can change "children" to "inner elements" if that's more comprehensible.

dirk-pieterse: I think so, or at least 'child elements'

PR #4550, opened 2023-01-16
DataGrid/PivotGrid: add encodeExecutableContent description

vladaskorohodova:


api-reference/_hidden/ExcelExportBaseProps/encodeExecutableContent.md
```diff
@@ -5,7 +5,9 @@ default: false
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies whether to encode potentially dangerous content when it is exported to a CSV file.
```
dmitry-eliseev-devexpress: The verb "to encode" can be understood in multiple ways. Yes, I understand that this verb is a part of the property name. But what does it mean here? Yes, you explain this below but your summary, your short description should explain what does this property actually do.

> Specifies whether to ...

I always advised against this pattern, for it sounds vague and disconnected. Our documentation should stick to the goal as much as possible, and we can hone this idea so this summary becomes razor-sharp.

> Specifies whether to encode potentially dangerous **content** when it is exported to a CSV file.

You do not distinguish between the exported data in general and potentially dangerous content. One also might think that everything that is exported is actually potentially dangerous content.

> ... when it is exported to a CSV file.

Avoidable passive voice.

Here's my take on this:

```suggestion
Specifies if the CSV export routine saves potentially dangerous content as plain text data.
```

Feel free to correct me if I misunderstood something...

api-reference/_hidden/ExcelExportBaseProps/encodeExecutableContent.md
```diff
@@ -5,7 +5,9 @@ default: false
...
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+Exported data can contain executable content such as formulas. This content makes it possible to execute malicious commands if a user opens a file in the spreadsheet application and confirms that the file can be loaded and commands executed.
```
dmitry-eliseev-devexpress: > Exported data can **contain** executable **content** such as formulas.

Contain content. You understand why this does not look good.

> This content **makes it possible** to execute malicious commands if a user opens a file in the spreadsheet application and confirms that the file **can be loaded and commands executed**.

More vague and indirect speech. Documentation should stick to the goal.

> if a user opens a file in **the** spreadsheet application

Since we haven't named this application, it literally can be **any** spreadsheet application.

Again, here's my take on this:

```suggestion
Exported spreadsheet documents can be unsafe because executable content (such as formulas) may include malicious code. A spreadsheet application can execute this code if a user opens such a file and confirms that the application can load and execute dynamic content.
```

api-reference/_hidden/ExcelExportBaseProps/encodeExecutableContent.md
```diff
@@ -5,7 +5,9 @@ default: false
...
+
+Enable this property to encode formulas into simple data when the content is exported to CSV.
```
dmitry-eliseev-devexpress: Exported where? To a file.

Exported how? In CSV format.

Besides, we've already explained everything above, so there is no need to repeat things over and over.

```suggestion
Enable this property to ensure that exported CSV files are safe for loading in third-party spreadsheet applications.
```

P.S. I wasn't aware of the fact that CSV files can include executable content. CSV is just "Comma-Separated Values" by its definition. One of the simplest plain text formats ever. Well, live and learn, I guess.

dmitry-eliseev-devexpress (commented):
Small topics may require a significant amount of effort when they explain non-trivial concepts. You need to hone the idea you need to convey so it serves the designed purpose perfectly. I know you can do it.

Documentation must stick to the goal as much as possible - be detailed but concise and easy to understand.

dirk-pieterse (approved):

dmitry-eliseev-devexpress (approved):

PR #4665, opened 2023-02-20
BarGauge: describe resolveLabelOverlapping

vladaskorohodova:


api-reference/10 UI Components/dxBarGauge/1 Configuration/resolveLabelOverlapping.md
```diff
@@ -6,6 +6,9 @@ default: 'hide'
 ---
 ---
 ##### shortDescription
-Specifies how the UI component should behave when [bar labels](/api-reference/10%20UI%20Components/dxBarGauge/1%20Configuration/label '/Documentation/ApiReference/UI_Components/dxBarGauge/Configuration/label/') overlap: hide certain labels or leave them overlapped.
+Specifies how the UI component should behave when [bar labels](/api-reference/10%20UI%20Components/dxBarGauge/1%20Configuration/label '/Documentation/ApiReference/UI_Components/dxBarGauge/Configuration/label/') overlap.
 
 ---
+If bar values in your BarGauge are close to each other, bar labels may overlap. In this case, use the **resolveLabelOverlapping** property to specify how the gauge must resolve overlapping. To hide certain labels, set this property to *'hide'*. The component automatically determines labels to hide. 
+
+The component can also shift labels from their positions. Set the **resolveLabelOverlapping** property to *'shift'* to use this functionality. If there is not enough space for all labels after they are shifted, the BarGauge hides labels with the smallest values.
```
Abadzhev: 1. I want to comment on paragraph structure. You want to introduce each paragraph with the main idea. And then you don't want to go beyond that idea in the that paragraph. In this text, the first two sentences are fine. They describe a general idea of overlapping labels and the property that controls their behavior. But then the third sentence describes a specific value. And another value, for some reason, has it's own paragraph. This is where I see somewhat unbalanced description. You probably want to describe each value in the same manner. And I would even suggest a list. 
2. I don't know if you need to use the word "bar" that many times in the first sentence: "bar values", "BarGauge", "bar labels". When I first read this, thought the sentence said "if bar values are close, bar values may overlap". Phrases are similar enough so I thought they are the same. But maybe it's just me :-) Anyway, I wouldn't repeat the word "bar" that many times. 
3. "If there is not enough space" - we try to avoid the sentence that use "there is", "there are", "it is", etc. It's best to find a clear pair of subject and verb. 
4. "The component automatically determines labels to hide." vs "hides labels with the smallest values". In one case, you leave the description vague - "automatically determines" (how?). In the next paragraph you are more specific about the algorithm. Can we say something more specific in the first case? 

```suggestion
If a BarGauge displays several values close to each other, the corresponding labels may overlap. If you want to avoid overlapped labels, use the **resolveLabelOverlapping** property to specify how the gauge adjusts label position or visibility. The following values are available: 

- *'hide'*. The gauge hides labels with the smallest values.
- *'shift'*. The gauge shifts labels. If the component's size is not big enough to accommodate all labels, the BarGauge hides labels with the smallest values.
```

api-reference/10 UI Components/dxBarGauge/1 Configuration/resolveLabelOverlapping.md
```diff
@@ -6,6 +6,16 @@ default: 'hide'
...
 ---
 ##### shortDescription
-Specifies how the UI component should behave when [bar labels](/api-reference/10%20UI%20Components/dxBarGauge/1%20Configuration/label '/Documentation/ApiReference/UI_Components/dxBarGauge/Configuration/label/') overlap: hide certain labels or leave them overlapped.
+Specifies how the UI component should behave when [bar labels](/api-reference/10%20UI%20Components/dxBarGauge/1%20Configuration/label '/Documentation/ApiReference/UI_Components/dxBarGauge/Configuration/label/') overlap.
 
 ---
+If a BarGauge displays several values close to each other, the corresponding labels may overlap. If you want to avoid overlapped labels, use the **resolveLabelOverlapping** property to specify how the gauge adjusts label position or visibility. The following values are available: 
+
+- *'hide'*    
+The gauge hides all labels with the smallest values except the first one.
```
albertov05: When you say "except the first one", do you mean that all values that are _smaller than the first label's value_ are hidden? Please clarify.

vladaskorohodova: I wanted to say that all smaller labels are hidden except the label with the smallest value. Like if you had [2, 3, 3, 3, 4, 5] labels, after the algorithm does its work it should be something like [2, 4, 5]. So all the smallest except the smallest of them all 😄

albertov05: Thanks for the clarification. How about the following?

```suggestion
The gauge hides labels with smaller values except for the label with the smallest value.
```

vladaskorohodova: Thanks, that's totally better!

PR #4949, opened 2023-04-25
Update '3rd-Party Template Engines' article

vladaskorohodova:


concepts/05 UI Components/zz Common/30 Templates/30 3rd-Party Template Engines.md
```diff
@@ -1,34 +1,15 @@
-You can use a 3rd-party template engine, but only with jQuery. DevExtreme supports the following template engines out of the box:
+You can use the <a href="http://mustache.github.io/" target="_blank">Mustache</a> 3rd-party template engine. Call the [DevExpress.setTemplateEngine(options)](/api-reference/50%20Common/utils/setTemplateEngine(options).md '/Documentation/ApiReference/Common/Utils/#setTemplateEngineoptions') method and implement functions that `compile` and `render` templates:
```
VasilyStrelyaev: > You can use the <a href="http://mustache.github.io/" target="_blank">Mustache</a> 3rd-party template engine.

In fact, this mechanism allows users to use any 3rd party template engine. We should be clear about that.

At the same time, we recommend only Mustache as the safest engine, that's why we don't list or provide samples for alternative engines. I think we can put it mildly, just by mentioning its advantage, like: "You can use a 3rd party template engine of your choice. Here is an example for Mustache, a CSP-compliant template engine...".

@dxvladislavvolkov

concepts/05 UI Components/zz Common/30 Templates/30 3rd-Party Template Engines.md
```diff
@@ -1,34 +1,15 @@
...
-- <a href="https://github.com/BorisMoore/jsrender" target="_blank">JsRender</a>
-- <a href="http://mustache.github.io/" target="_blank">Mustache</a>
-- <a href="http://twitter.github.io/hogan.js/" target="_blank">Hogan</a>
-- <a href="http://handlebarsjs.com/" target="_blank">Handlebars</a>
-- <a href="http://olado.github.io/doT/index.html" target="_blank">doT</a>
+    <!-- tab: index.js -->
+    DevExpress.setTemplateEngine({
+        compile: (element) => $(element).html(),
+        render: (template, data) => Mustache.render(template, data),
+    });
```
VasilyStrelyaev, line 11: Can we leave the template part?

```text
    <!-- tab: index.js -->
    DevExpress.setTemplateEngine({
        compile: (element) => $(element).html(),
        render: (template, data) => Mustache.render(template, data),
    });

    $(function() {
        $("#list").dxList({
            // ...
            itemTemplate: $("#itemTemplate")
        });
    })

    <!-- tab: index.html -->
    <div id="list"></div>
    <script type="text/html" id="itemTemplate">
        <!-- your Mustache template -->
    </script>
```

concepts/05 UI Components/zz Common/30 Templates/30 3rd-Party Template Engines.md
```diff
@@ -1,34 +1,15 @@
...
 
-To use one of them, pass its name to the [DevExpress.setTemplateEngine(name)](/api-reference/50%20Common/utils/setTemplateEngine(name).md '/Documentation/ApiReference/Common/Utils/#setTemplateEnginename') method:
+If you assign a function that returns a string to a template, the `compile` callback receives a text node as an argument. In this case, the `compile` callback needs to contain additional logic that determines if an argument is a template element or a text node:
```
VasilyStrelyaev: [minor] This is minor, but I'd like to be sure we are on the same page here.

The colon and the bullet list formatting here make an impression like the list describes the "additional logic" mentioned in the previous sentence.

However, that's not what we mean.

The users need to:
1. Implement additional logic to determine if an argument is a template element or a text node (we do not give details on it)
2. Based on the result of the previous step, either do `$(element).html()` or `$('<div>').append(element).html()`

VasilyStrelyaev (changes requested):

concepts/05 UI Components/zz Common/30 Templates/30 3rd-Party Template Engines.md
dxvladislavvolkov: I agree with Vasily. but I wouldn't mention CSP and and just write 'Here is an example for Mustache....'

VasilyStrelyaev (approved):

dxvladislavvolkov (approved):

dmitry-eliseev-devexpress (approved):

dirk-pieterse (approved):

PR #5011, opened 2023-05-12
DataGrid: update the calculateFilterExpression property

sergepilipchuk:


api-reference/_hidden/GridBaseColumn/calculateFilterExpression.md
```diff
@@ -202,6 +202,8 @@ If you [specify a custom header filter data source](/concepts/05%20UI%20Componen
 
 [/note]
 
+If you use the search panel, the DataGrid may invoke the **calculateFilterExpression** function multiple times for lookup columns. The first call is to filter the lookup's data source. Subsequent calls are to filter the DataGrid data source.
```
dirk-pieterse:
```suggestion
If you use the search panel, the DataGrid may invoke the **calculateFilterExpression** function multiple times for lookup columns. The first call is to filter the lookup's data source, and subsequent calls are to filter the DataGrid's data source.
```

dirk-pieterse (approved):

PR #5547, opened 2023-09-01
Tabs/TabPanel: describe `tabsPosition` and `orientation`

vladaskorohodova:


api-reference/10 UI Components/dxTabPanel/1 Configuration/tabsPosition.md
```diff
@@ -5,7 +5,15 @@ default: 'top'
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies tabs position relative to the panel.
```
Abadzhev: In noun phrases, the first noun (the one that acts as an adjective) is usually singular. You can say "a column collection" (even though you obviously mean multiple columns) or "60-day money back guarantee" (again, you obviously talk about multiple days).  Or you can have a property named "DataGridView.CellBorderStyle" (from Microsoft). It does say singular "cell", but it applies to all cells. 

My recommendation is to change the property name as well: `tabPosition`. If, for some reason, you don't like this, you can also go another route and call it `tabStripPosition`. Then you kind of unite all tabs into one "strip" and specify the position of this strip. But again, it's would not be "tabsStrip...", but "tabStrip...".  

```suggestion
Specifies tab position relative to the panel.
```

api-reference/10 UI Components/dxTabs/1 Configuration/orientation.md
```diff
@@ -5,7 +5,11 @@ default: 'horizontal'
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies Tabs orientation.
```
Abadzhev: The same comment applies as in the other member (singular vs plural in noun phrases). 

I would also use consistent capitalization.

```suggestion
Specifies tab orientation.
```

api-reference/10 UI Components/dxTabPanel/1 Configuration/tabsPosition.md
vladaskorohodova: I know that it's usually like that but the thing is it might be confusing if we call it `tabPosition`.
Both me and developers think that since we position all tabs, not a single tab, the name should reflect it. I mean, even the component is called Tabs. And TabPanel consists of Tabs and MultiView components. Our customers are used to Tabs in plural. That's why I don't like these suggestions. But I can't come up with something better than `tabsPosition`.

Abadzhev: I'm not sure you will confuse anyone by using the name "tabPosition". But the world is not going to end if you keep it named "tabsPosition", I guess :-) Truth be told, I can find similar properties in MS docs too. Here's an example: 

https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.datagridview.columnheadersheight?view=windowsdesktop-7.0

But we do step into the area of questionable grammar if we name properties like that :-) "tabStripPosition" would be a good compromise.  

I need to share an example of a tabPosition property, though: 

https://developer.apple.com/documentation/appkit/nstabview
https://developer.apple.com/documentation/appkit/nstabview/tabposition

Anyway - at this point, I shared all my thoughts with you. Go with whatever you guys decide as a team. It's not that big of a deal, of course.

api-reference/10 UI Components/dxTabPanel/1 Configuration/tabsPosition.md
```diff
@@ -5,7 +5,15 @@ default: 'top'
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies tab position relative to the panel.
```
dirk-pieterse, line 8:
```suggestion
Specifies the tab's position relative to the panel.
```

api-reference/10 UI Components/dxTabs/1 Configuration/orientation.md
```diff
@@ -5,7 +5,11 @@ default: 'horizontal'
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies the component orientation.
```
dirk-pieterse:
```suggestion
Specifies the component's orientation.
```

api-reference/10 UI Components/dxTabs/1 Configuration/orientation.md
```diff
@@ -5,7 +5,11 @@ default: 'horizontal'
...
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+The following orientation modes are available:
```
dirk-pieterse:
```suggestion
The following orientations are available:
```

dirk-pieterse (approved):

api-reference/10 UI Components/dxTabs/1 Configuration/orientation.md
dirk-pieterse: or "Specifies component orientation"

dirk-pieterse (approved):

api-reference/10 UI Components/dxTabs/1 Configuration/orientation.md
vladaskorohodova:
```suggestion
Specifies component orientation.
```

PR #5579, opened 2023-09-18
Form/Editors: describe `isDirty`, `clear` and `reset`

vladaskorohodova:


api-reference/10 UI Components/Editor/1 Configuration/isDirty.md
```diff
@@ -5,7 +5,146 @@ default: false
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies whether the component's current value differs from the initial value.
+
+##### readonly
+
+---
+This property is a readonly flag. You can use it to check whether the editor was changed.
```
Abadzhev: I would be a bit more pedantic here and use a hyphen for "read-only". I'm sure some resources may now write it as one word, but it just seems wrong to me :-) 

"editor was changed" is not an entirely correct phrase. If you resize an editor, or change its color, you're changing the editor. Let's incorporate the word "value" here to be more specific. I'm also not sure that "changed" or "was changed" differ in this context. 

```suggestion
This property is a read-only flag. You can use it to check whether the editor value changed.
```

api-reference/10 UI Components/Editor/3 Methods/reset(value).md
```diff
@@ -3,10 +3,10 @@ id: Editor.reset(value)
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Resets the [value](/api-reference/10%20UI%20Components/Editor/1%20Configuration/value.md '{basewidgetpath}/Configuration/#value') property to the value passed as an argument.
 
 ##### param(value): any
-<!-- Description goes here -->
+Specifies new value.
```
Abadzhev:
```suggestion
Specifies the new value.
```

api-reference/10 UI Components/dxForm/1 Configuration/isDirty.md
```diff
@@ -5,7 +5,12 @@ default: false
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies whether editors current values differ from the initial values.
```
Abadzhev: "editors current values" reads wrong to me. First of all, "editor" needs to act as an adjective for the word "value". And it doesn't matter if you talk about a value (singular) or values (plural). The modifier (adjective) stays the same - "editor". You can say "specifies whether the editor value" or "specifies whether editor values". 

Now the word "current" should precede this noun phrase: "current editor values". Otherwise you'd have to switch to possessive - "editors' current values". And it's best to avoid plural possessives. 

Finally, you don't need an article before "initial values". It's plural and not really a specific reference. 

```suggestion
Specifies whether current editor values differ from initial values.
```

api-reference/10 UI Components/dxForm/1 Configuration/isDirty.md
```diff
@@ -5,7 +5,12 @@ default: false
...
+
+##### readonly
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+This property is a readonly flag. You can use it to check whether the editors in Form were changed.
```
Abadzhev:
```suggestion
This property is a read-only flag. You can use it to check whether any editor values changed on this Form. 
```

api-reference/10 UI Components/dxForm/3 Methods/clear().md
```diff
@@ -3,7 +3,7 @@ id: dxForm.clear()
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Resets the editor values to their default values.
```
Abadzhev:
```suggestion
Resets editor values to their default values.
```

api-reference/10 UI Components/dxForm/3 Methods/reset(editorsData).md
```diff
@@ -3,10 +3,142 @@ id: dxForm.reset(editorsData)
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Resets the editors values to the values passed as an argument. Resets all other values to initial values.
```
Abadzhev: I rewrote this paragraph, because right now the sentences are not linked. If you read the first, you don't anticipate that a second will come. You think that ALL editors are already updated. But then you read about OTHER editors. I wanted to make sure that the reader can easily follow our thoughts - from the start they understand that the argument may or may not contain a value for a specific editor. 

```suggestion
Resets editor values. Every editor attempts to obtain its new value from the argument list. If an editor is not included in the list, that editor resets to its initial value.
```

api-reference/10 UI Components/dxForm/3 Methods/reset(editorsData).md
```diff
@@ -3,10 +3,142 @@ id: dxForm.reset(editorsData)
...
 
 ##### param(editorsData): Object
-<!-- Description goes here -->
+New editors data.
```
Abadzhev: Again - even if "data" is plural, its modifier/adjective should be singular: "editor data". In other words "editor data" can mean data for multiple editors, no problems with that. But since the description above uses the word "values" instead of "data", I decided to change it a bit. 

```suggestion
New values for editors.
```

api-reference/10 UI Components/dxForm/1 Configuration/isDirty.md
```diff
@@ -5,7 +5,12 @@ default: false
...
 ##### shortDescription
-<!-- Description goes here -->
+Specifies whether current editor values differ from initial values.
+
+##### readonly
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+This property is a read-only flag. You can use it to check whether any editor values changed on this Form.
```
dirk-pieterse:
```suggestion
This property is a read-only flag. You can use it to check if any editor value changed on this Form. 
```

api-reference/10 UI Components/Editor/1 Configuration/isDirty.md
```diff
@@ -5,7 +5,146 @@ default: false
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Specifies whether the component's current value differs from the initial value.
+
+##### readonly
+
+---
+This property is a read-only flag. You can use it to check whether the editor value changed.
```
dirk-pieterse:
```suggestion
This property is a read-only flag. You can use it to check if the editor value changed.
```

dirk-pieterse (approved):

PR #5899, opened 2023-12-26
Validator: update Email rule

vladaskorohodova:


api-reference/10 UI Components/dxValidator/8 Validation Rules/EmailRule/EmailRule.md
```diff
@@ -13,5 +13,9 @@ A validation rule that demands that the validated field match the Email pattern.
     href: "https://js.devexpress.com/Demos/WidgetsGallery/Demo/Validation/Overview/"
 }
 
+The Email pattern in DevExtreme components is the following:
```
albertov05: Is it possible to turn this line into an instructional statement? For example:

`Use the following email pattern in DevExtreme components:`

Also, regarding Line 9 above:
"A validation rule that demands that the validated field match the Email pattern."

Can we change "demands" to "requires"?

PR #6021, opened 2024-02-15
DataGrid: update Column and Row Indexes topic

vladaskorohodova:


flagmanAndrew (approved):

arminal:
Please change this:
"The default index, *-1*, means that no row is focused."
to this:
"The default index of *-1* indicates that no row is focused."

PR #6073, opened 2024-02-27
DropDownButton: describe `template`

vladaskorohodova:


api-reference/10 UI Components/dxDropDownButton/1 Configuration/template.md
```diff
@@ -4,22 +4,191 @@ type: template
...
-<!-- Description goes here -->
+The button's text.
 
 ##### param(contentElement): DxElement
-<!-- Description goes here -->
+#include common-ref-elementparam with { element: "button content" }
 
 ##### return: String | Element | jQuery
-<!-- Description goes here -->
+A template name or container.
```
Abadzhev: I'm not sure about this one. Seems like this is the return value, so I probably wouldn't use "a" as an article. You would say "a template name" for a parameter... You can pass any template name. But if it's the return value, then this should be the result of some calculation or some query... So I'd likely say "the". Maybe something like this: 

```suggestion
The template name or the template's container element.
```

api-reference/10 UI Components/dxDropDownButton/1 Configuration/template.md
```diff
@@ -4,22 +4,191 @@ type: template
...
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+The following code sample creates a DropDownButton featuring multiline text for the base button.
```
Abadzhev:
```suggestion
The following code sample creates a DropDownButton with multiline text in the base button.
```

dirk-pieterse (approved):

PR #6509, opened 2024-08-05
DataGrid: update editCellTemplate

vladaskorohodova:


arminal (commented):
Line 7, change to:
Specifies a custom template for data cells in an editing state.
Line 13, change to:
Cell properties.
Line 26, change to:
Cell row data.
Line 32, change to:
Cell rows.
Line 39, change to:
Row types.
Line 42, change to:
A method called to change cell values; optionally, you can also call it to change the displayed value after the editor's value is changed.
Line 45, change to:
[note] In batch [editing mode](/api-reference/10%20UI%20Components/dxDataGrid/1%20Configuration/editing/mode.md '{basewidgetpath}/Configuration/editing/#mode'), confirm that the value is changed before calling this method to ensure correct cell highlighting.
Line 67, change to:
Use **editCellTemplate** to replace the default editor. In the template, specify replacement editor appearance and behavior.
Line 82, change to:
[Dynamically change editor properties in the editing state]

PR #6541, opened 2024-08-14
Update viz.refreshTheme() method description

vladaskorohodova:


api-reference/50 Common/utils/viz/refreshTheme().md
```diff
@@ -8,6 +8,8 @@ export: refreshTheme
 Refreshes the current theme and palette in all data visualization UI components on the page.
 
 ---
+The **refreshTheme()** method calls [refresh()](/Documentation/ApiReference/UI_Components/dxChart/Methods/#refresh) to reload data and update the UI of all data visualization components after applying a new theme.
```
albertov05, line 11: Just to make sure my understanding is correct, the **refreshTheme()** method calls another method - **refresh()**?

vladaskorohodova: Yes, the source code of **refreshTheme()** calls **refresh()**

albertov05: Okay, thanks for the clarification

PR #6809, opened 2024-11-19
TreeView: add a note about rootValue

vladaskorohodova:


api-reference/10 UI Components/dxTreeView/1 Configuration/dataStructure.md
```diff
@@ -8,7 +8,7 @@ default: 'tree'
 Notifies the UI component of the used data structure.
```
albertov05: I'd like to clarify "the used data structure". Are you talking about the data structure that is currently being used? If so, then would it be better to say:

`Notifies the UI component of the data structure in use.`

api-reference/10 UI Components/dxTreeView/1 Configuration/dataStructure.md
```diff
@@ -8,7 +8,7 @@ default: 'tree'
...
 
 ---
-The UI component expects that data has a hierarchical structure where each data item contains a `text`, a unique `id`, and an `items` array if the data item nests other data items:
+The UI component expects that data has a hierarchical structure (`dataStructure: 'tree'`). Each data item contains a `text`, a unique `id`, and an `items` array if the data item includes sub-items:
```
albertov05: In this context, can we refer to `text`, `id` and `items` as "properties"? If so, would it be technically correct to say something like:

``The UI component expects that data has a hierarchical structure (`dataStructure: 'tree'`). Each data item contains `text` and `id` properties, and an `items` property (containing an array) if the data item includes sub-items:``

Alternatively, we could avoid using the inline code style and say something like:

``The UI component expects that data has a hierarchical structure (`dataStructure: 'tree'`). Each data item contains text, a unique id, and an items array if the data item includes sub-items:``

Let me know what you think.

api-reference/10 UI Components/dxTreeView/1 Configuration/dataStructure.md
```diff
@@ -27,6 +27,8 @@ The UI component expects that data has a hierarchical structure where each data
         ]
     }];
 
+[note] A data item's [id](/Documentation/ApiReference/UI_Components/dxTreeView/Configuration/items/#id) should not match [rootValue](/Documentation/ApiReference/UI_Components/dxTreeView/Configuration/#rootValue) in the 'tree' **dataStructure**.
+
 #include btn-open-demo with {
     href: "https://js.devexpress.com/Demos/WidgetsGallery/Demo/TreeView/HierarchicalDataStructure/"
 }
```
albertov05, line 34: Regarding line 34 below:  
``If data has a plain structure, set the **dataStructure** property to *"plain"*. In this case, each data item should have a `text`, a unique `id`, and a `parentId`. For root items, `parentId` should be equal to **0** or **undefined**:``

Please see my comment above.

api-reference/10 UI Components/dxTreeView/1 Configuration/rootValue.md
```diff
@@ -8,3 +8,4 @@ default: 0
 Specifies the [parent ID](/api-reference/10%20UI%20Components/dxTreeView/1%20Configuration/parentIdExpr.md '/Documentation/ApiReference/UI_Components/dxTreeView/Configuration/#parentIdExpr') value of the root item.
 
 ---
+[note] A data item's [id](/Documentation/ApiReference/UI_Components/dxTreeView/Configuration/items/#id) should not match **rootValue** in the 'tree' [dataStructure](/Documentation/ApiReference/UI_Components/dxTreeView/Configuration/#dataStructure).
```
albertov05: Instead of "dataStructure", can we say "data structure"?

PR #7174, opened 2025-04-02
Map: update routes[].mode

vladaskorohodova:


Zedwag (approved):

api-reference/10 UI Components/dxMap/1 Configuration/routes/mode.md
```diff
@@ -8,4 +8,28 @@ default: 'driving'
...
+    - `truck`
+    - `van`
+
+- [Bing](https://learn.microsoft.com/en-us/bingmaps/v8-web-control/modules/directions-module/routemode-enumeration)
+    - `driving`
+    - `transit`
+    - `truck`
+    - `walking`
+
+[note] DevExtreme Map allows access to part of the provider’s API as is. We do not guarantee correct application of the route mode if the string does not match the expected values or the specific configuration for the provider.
```
Abadzhev: I have a couple of questions about this. 

> We do not guarantee correct application of the route mode if the string does not match the expected values...

I guess I understand that. If I use Bing, and I specify "bicycle", then there will be issues. Looks like Bing doesn't support that value (Azure does). But I think direct access to provider's API has nothing to do with this (the previous sentence). Whether I have direct access to provider's API or not, I just specified an incorrect value. I'll get an error probably, or just some unexpected behavior. 

> We do not guarantee correct application of the route mode if the string does not match ... the specific configuration for the provider. 

If I understand this correctly, I can use provider's API to change the configuration. And then some route mode value (say, "transit"), will no longer work, because of current configuration? If I'm wrong about this, can you please give me an example of configuration and conflicting route mode?  

My current edit would be something like that... But I'm not sure: 

```suggestion
[note] If the route doesn't match the mode you specified, consider the following possible causes: 

* Not every provider supports all possible route modes. Make sure that you specify a supported mode. 
* When assigning a string, check that you specify a provider-specific spelling (compare Google's `BICYCLING` and Azure's `bicycle`). 
* DevExtreme Map allows partial access provider’s API. If you changed configuration, the provider may no longer support the route mode you specified. 
```

api-reference/10 UI Components/dxMap/1 Configuration/routes/mode.md
```diff
@@ -8,4 +8,28 @@ default: 'driving'
 Specifies a transportation mode to be used in the displayed route.
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+Choose a predefined route mode, such as `'driving'` or `'walking'`, or select one supported by your provider:
```
Abadzhev: So these two modes that you mention - `driving` and `walking` - can I specify them regardless of the provider? 

The lists below show that Azure doesn't have a `walking` mode. It's `pedestrian`. Can I still specify `walking`? 

And if I understand the above correctly... What other "predefined route modes" are available? Why do you only specify a couple of examples? I think a complete list would be useful.

vladaskorohodova: There are only 2 predefined modes: 'walking' and 'driving'. You can specify them regardless of the provider (Map converts them to the necessary modes, e.g. you specify "walking" for Azure and it works).

api-reference/10 UI Components/dxMap/1 Configuration/routes/mode.md
vladaskorohodova: 1. Previously, all the calculations for the mode were built-in and a user couldn't pass anything besides "driving" and "walking" (if they passed anything else the string just changed to "driving" by default). Now we allow users direct access to provider's API through Map API. We just check if it's "walking" or "driving" and if it's not - we pass the route as is.
2. You are right. For instance, in Google the "transit" mode only works if you specify a route out of 2 points. If you pass a route out of 3+ points, the route won't be rendered. But this error is specific to Google maps, so we do not guarantee that this will work on our side too.

api-reference/10 UI Components/dxMap/1 Configuration/routes/mode.md
Abadzhev: Then I would be specific that these two predefined modes are the only two predefined modes available. In other words, I wouldn't use the phrase "such as", as it may mean that there are additional predefined modes. 

```suggestion
Regardless of the current provider, you can specify one of two predefined modes: `'driving'` or `'walking'`. You can also select a provider-specific mode:
```

api-reference/10 UI Components/dxMap/1 Configuration/routes/mode.md
Abadzhev: So it looks like my comment mostly works? I would edit it a little bit, like this: 

```suggestion
If the route doesn't render as expected, consider the following possible causes: 

* Not every provider supports all possible route modes. Make sure that you specify a supported mode. 
* Check that you specify a provider-specific spelling (compare Google's `BICYCLING` and Azure's `bicycle`). 
* DevExtreme Map allows partial access provider’s API. If you changed configuration, the provider may no longer support the route mode you specified. 
```

(BTW, I'm not sure if I can wrap it all into a note - what the syntax would be... So I removed the NOTE syntax for now. I assume you'll bring it back.)

api-reference/10 UI Components/dxMap/1 Configuration/routes/mode.md
```diff
@@ -8,4 +8,36 @@ default: 'driving'
 Specifies a transportation mode to be used in the displayed route.
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+Regardless of the current provider, you can specify one of two predefined modes: `'driving'` or `'walking'`. You can also select a provider-specific mode:
```
dirk-pieterse:
```suggestion
You can choose between two predefined modes— `'driving'` or `'walking'`—regardless of the current provider. Additionally, provider-specific modes are available:
```

api-reference/10 UI Components/dxMap/1 Configuration/routes/mode.md
```diff
@@ -8,4 +8,36 @@ default: 'driving'
...
+
+- [Bing](https://learn.microsoft.com/en-us/bingmaps/v8-web-control/modules/directions-module/routemode-enumeration)
+    - `driving`
+    - `transit`
+    - `truck`
+    - `walking`
+
+[note]
+
+If the route doesn't render as expected, consider the following possible causes:
```
dirk-pieterse:
```suggestion
If the route does not render as expected, consider the following: 
```

dirk-pieterse (approved):

PR #7391, opened 2025-06-03
Validator: add multiple messages code snippet

arman-boyakhchyan:


api-reference/10 UI Components/dxValidator/8 Validation Rules/CustomRule/validationCallback.md
arminal, line 1: Line 10 change to:
An object that defines validation parameters.
Line 16 change to:
Current row data. Exists only when you validate DataGrid or TreeList cell values.
Line 34 change/edit to: 
The following code sample shows a **validationCallback** example. The function (which function, specify) accepts a number, and returns **true** if the number is even and **false** if it is odd:

vladaskorohodova (approved):

PR #7430, opened 2025-06-09
CardView: Add API code snippets

arman-boyakhchyan:


api-reference/10 UI Components/dxCardView/1 Configuration/cardCover.md
```diff
@@ -8,4 +8,69 @@ inheritsType: CardCover
...
+
+---
+
+##### jQuery
+
+    <!-- tab: index.js -->
+    $(#dxCardView).dxCardView({
+        // ...
+        cardCover: {
+            imageExpr: "/images/image.jpg",
```
pomahtri: `imageExpr` and `altExpr` are selectors, not end-values that will be used for final value inside image
They are either functions like `(data) => data.myImageUrl` or strings which point to field inside data object. E.g. `imageExpr: "myImageUrl"` and `imageExpr: (data) => data.myImageUrl"` are the same
So provided sample will not work, let's change expr to func

api-reference/10 UI Components/dxCardView/9 Types/ColumnProperties/calculateFieldValue.md
```diff
@@ -16,3 +16,79 @@ A field's custom value.
...
+        <dxi-card-view-column
+            caption="Location"
+            [calculateFieldValue]="calculateLocation"
+        ></dxi-card-view-column>
+    </dx-card-view>
+
+    <!-- tab: app.component.ts -->
+    // ...
+    export class AppComponent {
+        calculateLocation(cardData: cardDataObject): string {
```
pomahtri: I'd suggest to remove `cardDataObject` type from here, customer might think that it is devextreme's type and will not find it. Or we can replace its name to smth like `Employee` to highlight that it is customer-level type

pomahtri (approved):

PR #7498, opened 2025-06-24
Grids: Add Lookup dataSource function note

arman-boyakhchyan:


api-reference/_hidden/GridBaseColumn/lookup/dataSource.md
```diff
@@ -36,6 +36,97 @@ An [ArrayStore](/api-reference/30%20Data%20Layer/ArrayStore '/Documentation/ApiR
 - **Function**      
 A function that returns one of the above.
 
+    [note]
+    
+    When the component is initialized, it calls the **dataSource** function with an empty **options** object. To avoid execution errors, check the truthiness of **options** properties before accessing them. The following code snippet demonstrates how you can check the truthiness of **options**.**data** before using the property in a filter expression:
```
dirkpieterse: `truthiness`? Can you explain what it means here?

arman-boyakhchyan: It simply means that the properties in question are not `undefined` or `null`. I thought about writing something like "Check that **options** properties are not `undefined` or `null` before accessing them." but ultimately decided to go with "truthiness" to make the sentence a touch more eloquent.

dirkpieterse: `truthiness` is the quality of seeming or being perceived as true, even if not necessarily true. I've never seen it used in this context.

arman-boyakhchyan: I thought it was a well-known term, it is even listed in the [MDN docs](https://developer.mozilla.org/en-US/docs/Glossary/Truthy). If you feel it is going to be confusing for a significant portion of our readers, I am open to replacing it with something simpler.

dirkpieterse: I would go with something like "Ensure option properties are defined before using them"

api-reference/_hidden/GridBaseColumn/lookup/dataSource.md
```diff
@@ -36,6 +36,97 @@ An [ArrayStore](/api-reference/30%20Data%20Layer/ArrayStore '/Documentation/ApiR
 - **Function**      
 A function that returns one of the above.
 
+    [note]
+    
+    When the component is initialized, it calls the **dataSource** function with an empty **options** object. To avoid execution errors, check that **options** properties are defined before accessing them. The following code snippet demonstrates how you can check if **options**.**data** is defined before using the property in a filter expression:
```
dirkpieterse:
```suggestion
    When the component is initialized, it calls the **dataSource** function with an empty **options** object. To avoid execution errors, check that **options** properties are defined before accessing them. The following code snippet demonstrates how to check if **options**.**data** is defined before using the property in a filter expression:
```

vladaskorohodova (approved):

PR #7605, opened 2025-07-09
Charts: Update range property topics

arman-boyakhchyan:


api-reference/10 UI Components/dxRangeSelector/1 Configuration/value/value.md
```diff
@@ -10,12 +10,194 @@ notUsedInTheme:
...
+    ##### Angular
+
+        <!-- tab: app.component.html -->
+        <dx-range-selector [value]="{ startValue: 50, endValue: 70 }" ></dx-range-selector>
+        <!-- Open-ended range -->
+        <dx-range-selector [value]="{ startValue: null, endValue: 70 }" ></dx-range-selector>
+
+    ##### Vue
+
+    The following code snippet demonstrates **value** defined as a `<DxRangeSelector>` property:
```
albertov05: For my understanding, if startValue and endValue are fields, what exactly is "value"? Is it an attribute?

api-reference/10 UI Components/dxRangeSelector/1 Configuration/value/value.md
```diff
@@ -10,12 +10,194 @@ notUsedInTheme:
...
+    ##### React
+
+    The following code snippet demonstrates **value** defined as a `<RangeSelector>` property:
+
+        <!-- tab: App.tsx -->
+        <RangeSelector value={{ startValue: null, endValue: 70 }} />
+        <!-- Open-ended range -->
+        <RangeSelector value={{ startValue: null, endValue: 70 }} />
+
+    The following code snippet demonstrates value defined as a `<Value>` selector:
```
albertov05: Should bold font be applied to "value" (similar to the lines above)?

api-reference/10 UI Components/dxRangeSelector/1 Configuration/value/value.md
```diff
@@ -10,12 +10,194 @@ notUsedInTheme:
...
+            <Value startValue="50" length="20" />
+        </RangeSelector>
+        <!-- or -->
+        <RangeSelector>
+            <Value endValue="70" length="20" />
+        </RangeSelector>
+
+    ---
+
+    If you specify only **length**, the component implements the largest scale value as **endValue**.
```
albertov05: What do you mean by "implements" in this context? For example, would it be technically correct to say something like:

`If you specify only **length**, the component sets the largest scale value to the **endValue** value.`

includes/dataviz-ref-vizrangevariants.md
```diff
@@ -4,42 +4,240 @@ This property accepts one of the following:
...
+            <${{angularSelector}} [${{objectName}}]="{ startValue: 50, endValue: 70 }" ></${{angularSelector}}>
+        </${{angularComponentName}}>
+        <!-- Open-ended range -->
+        <${{angularComponentName}}>
+            <${{angularSelector}} [${{objectName}}]="{ startValue: null, endValue: 70 }" ></${{angularSelector}}>
+        </${{angularComponentName}}>
+
+    ##### Vue
+
+    The following code snippet demonstrates **${{objectName}}** defined as a `<Dx${{parentSelector}}>` property:
```
albertov05: For my understanding, is `<Dx${{parentSelector}}>` a placeholder, or is this the exact text that will appear on the documentation page?

includes/dataviz-ref-vizrangevariants.md
```diff
@@ -4,42 +4,240 @@ This property accepts one of the following:
...
+        <!-- tab: App.tsx -->
+        <${{componentName}}>
+            <${{parentSelector}} ${{objectName}}={{ startValue: null, endValue: 70 }} />
+        </${{componentName}}>
+        <!-- Open-ended range -->
+        <${{componentName}}>
+            <${{parentSelector}} ${{objectName}}={{ startValue: null, endValue: 70 }} />
+        </${{componentName}}>
+
+    The following code snippet demonstrates ${{objectName}} defined as a `<${{objectSelector}}>` selector:
```
albertov05: See my comment above.

api-reference/10 UI Components/dxRangeSelector/1 Configuration/value/value.md
arman-boyakhchyan: Yeah on second thought attribute is far more accurate in this context, I will update the text.

api-reference/10 UI Components/dxRangeSelector/1 Configuration/value/value.md
arman-boyakhchyan: Yes, seems like I missed bold formatting in a couple of spots. Thank you for noticing!

api-reference/10 UI Components/dxRangeSelector/1 Configuration/value/value.md
arman-boyakhchyan: If you specify only **length**, the component will behave as if **endValue** is specified as the largest scale/axis value.  I kind of wanted to word it as if the component adds the **endValue** property to this configuration object. I aggree with your suggestion though, as it is more straightforward. I would only like to suggest rewording it to avoid the "value to the endValue value" bit at the end:
```suggestion
    If you specify only **length**, the component defines the **endValue** property as the largest scale value. 
```

includes/dataviz-ref-vizrangevariants.md
arman-boyakhchyan: It's actually a variable that is defined in the docs that implement this `#include`, all of which have been updated in this PR as well. These variable values will appear in the docs anywhere there is a ${{variableName}} syntax.

api-reference/10 UI Components/dxRangeSelector/1 Configuration/value/value.md
albertov05: Thank you!

api-reference/10 UI Components/dxRangeSelector/1 Configuration/value/value.md
albertov05: Sounds good, thank you.

includes/dataviz-ref-vizrangevariants.md
albertov05: Okay, I see. The main reason I asked is that if they are variables, it's difficult to judge whether the appropriate indefinite article is "a" or "an" - without actually seeing these variables.

arman-boyakhchyan: Oh I had not thought about that at all. Yes, in the React docs there are instances where the "a" article is not appropriate. I'll update the text in both React and Vue versions of the docs for consistency. Thank you for noticing this!

albertov05: Thank you!

vladaskorohodova (approved):

PR #8310, opened 2026-01-05
Vue & React: Clarify Inline Object Warning Wording

arman-boyakhchyan:


artem-kurchenko (approved):

concepts/50 React Components/40 Component Configuration Syntax/12 Properties of the Object Type.md
```diff
@@ -72,9 +72,9 @@ Object type properties that depend on other properties' values are not implement
         }
     }
 
-[important] We recommend that you declare the object outside the configuration component to prevent possible issues caused by unnecessary re-rendering.
+[important] To avoid issues caused by excess re-renders, do not define inline objects in markup. Use configuration components or assign variables to object properties instead.
```
dirkpieterse:
```suggestion
[important] To avoid issues caused by excessive re-rendering, do not define inline objects in markup. Use configuration components or assign variables to object properties instead.
```

PR #8410, opened 2026-02-03
ODataStore: Add processDatesAsUtc Desc.

arman-boyakhchyan:


api-reference/30 Data Layer/ODataStore/1 Configuration/processDatesAsUtc.md
```diff
@@ -5,7 +5,10 @@ default: false
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Configures whether to ignore specified time zones for *"Date"* and *"DateTimeOffset"* [field types](/Documentation/ApiReference/Data_Layer/ODataStore/Configuration/#fieldTypes) (in requests and attached components).
```
Abadzhev: If it's a simple value type - Boolean, integer, string, and so on - I would just say "Specifies". The word "configures" seems more appropriate if you use an object with multiple settings. 

And I see you already have the word "specified" (probably that's why you decided to go with a synonym in the first place). So I want to simplify this short description a little bit - move details to longer remarks section, remove parentheses, make it more active:
 
```suggestion
Specifies whether requests and attached components ignore local time zone when treating date-time field values. 
```

api-reference/30 Data Layer/ODataStore/1 Configuration/processDatesAsUtc.md
```diff
@@ -5,7 +5,10 @@ default: false
...
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+When enabled, ODataStore ignores specified time zones and processes date values as UTC (sets the [time zone designator](https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators) in an ISO 8601 string to "Z"). When disabled (default), the store uses the following time zone offsets to convert dates to UTC:
```
Abadzhev: I'm adding a part with field value types (the part I removed from short description). And I also add a bit more structure. I was thinking about a list (for `true` and `false`)... But since you already have another list, and descriptions are quite lengthy, I'm going with separate paragraphs. 

```suggestion
This option specifies how to treat values in *"Date"* and *"DateTimeOffset"* [field types](/Documentation/ApiReference/Data_Layer/ODataStore/Configuration/#fieldTypes). 

If the option is `true`, ODataStore ignores specified time zones and processes date values as UTC (sets the [time zone designator](https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators) in an ISO 8601 string to "Z"). 

If the option is `false` (default), the store uses the following time zone offsets to convert dates to UTC:
```

api-reference/30 Data Layer/ODataStore/1 Configuration/processDatesAsUtc.md
arman-boyakhchyan: I would like to slightly modify your suggestion, as it is not only local time zones that are ignored, but also those specified in date-time field values themselves:

```suggestion
Specifies whether requests and attached components ignore time zones when treating date-time field values.
```

api-reference/30 Data Layer/ODataStore/1 Configuration/processDatesAsUtc.md
arman-boyakhchyan: That works perfectly, thank you!

api-reference/30 Data Layer/ODataStore/1 Configuration/processDatesAsUtc.md
Abadzhev: Yes, sounds good.

dirkpieterse (approved):

mpreyskurantov (approved):

PR #8610, opened 2026-03-25
React State Mgmt.: Add Default dataSource Limitation Note

arman-boyakhchyan:


markallenramirez (approved):

concepts/50 React Components/20 State Management/7 Uncontrolled Mode.md
```diff
@@ -14,3 +14,5 @@ To specify an initial property value in this mode, add the `default` prefix to p
             />
         );
     };
+
+[note] You cannot use this approach to specify an initial `dataSource`.
```
dirkpieterse: Can there be multiple initial data sources?

arman-boyakhchyan: What I meant here is that you cannot use this approach to specify a component's initial data source:
```suggestion
[note] You cannot use this approach to specify a component's initial `dataSource`.
```

PR #8633, opened 2026-04-02
Scheduler: add Occurrences topic

vladaskorohodova:


Tucchhaa (approved):

concepts/05 UI Components/Scheduler/035 Occurrences.md
```diff
@@ -0,0 +1,46 @@
+**Occurrence** is an [appointment](/Documentation/Guide/UI_Components/Scheduler/Appointments/Appointment_Types/) instance that falls within a specified time interval. A non-recurring appointment maps to one occurrence. A recurring appointment can generate multiple occurrences.
```
Abadzhev: In this case, we seem to treat the word "occurrence" as a regular word. So we should apply regular articles. (Doesn't sound like Occurrence is a proper noun.)

```suggestion
An **Occurrence** is an [appointment](/Documentation/Guide/UI_Components/Scheduler/Appointments/Appointment_Types/) instance that falls within a specified time interval. A non-recurring appointment maps to one occurrence. A recurring appointment can generate multiple occurrences.
```

concepts/05 UI Components/Scheduler/035 Occurrences.md
```diff
@@ -0,0 +1,46 @@
...
+
+[note] Occurrences are produced for the requested interval only.
```
Abadzhev: This is a little bit unclear. Since it's passive, I'm not sure what you mean by "produced". Do you mean the control produces them when I'm looking at my calendar? Or do you mean that the method you mention below only generates a limited set of appointments. 

Either way, what is the practical consequence for me? Yeah - if I think about it, the control probably will not generate recurring appointments for every Wednesday, all the way into eternity. 

Also - when I create a recurring appointment, Outlook makes me specify start and end dates anyway. So there is always a limit. 

So let's clarify this line - what exactly does this mean and what are practical implications for me as a developer?

concepts/05 UI Components/Scheduler/035 Occurrences.md
```diff
@@ -0,0 +1,46 @@
...
+
+To get occurrences, use the [getOccurrences(startDate, endDate, appointments)](/Documentation/ApiReference/UI_Components/dxScheduler/Methods/#getOccurrencesstartDate_endDate_appointments) method. Specify the interval boundaries with `startDate` and `endDate` and pass the appointments array as the third argument.
```
Abadzhev: I don't want repeat "get occurrences" and `getOccurrences`. 

```suggestion
To obtain an occurrence list, call the [getOccurrences(startDate, endDate, appointments)](/Documentation/ApiReference/UI_Components/dxScheduler/Methods/#getOccurrencesstartDate_endDate_appointments) method. Pass `startDate`, `endDate`, and the appointment array.
```

concepts/05 UI Components/Scheduler/035 Occurrences.md
```diff
@@ -0,0 +1,46 @@
...
+
+Apply occurrences in the following scenarios:
```
Abadzhev:
```suggestion
You may need this method in the following scenarios:
```

concepts/05 UI Components/Scheduler/035 Occurrences.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Detect Overlapping Appointments
+
+Use this function to prevent appointment time conflicts when users create or edit appointments. The function checks if a candidate interval intersects with any occurrence in the selected day.
```
Abadzhev: When you say "this", I assume that you point to something that has been mentioned previously. But I guess you mean the function declared below. 

```suggestion
Use the following function to prevent appointment time conflicts when users create or edit appointments. The function checks if a candidate interval intersects with any occurrence in the selected day.
```

concepts/05 UI Components/Scheduler/035 Occurrences.md
```diff
@@ -0,0 +1,46 @@
...
+
+        const allAppointments = scheduler.getDataSource().items();
+
+        const occurrences = scheduler.getOccurrences(dayStart, dayEnd, allAppointments);
+        return occurrences.some((item) =>
+             item.startDate < candidate.endDate && candidate.startDate < item.endDate
+        );
+    }
+
+### Prepare Exports for a Visible Interval
```
Abadzhev: I'm not sure why you mention "visibility". The code simply sets start date and end date. I can export appointments from one week, while looking at a month view (an entire month will be visible). 

In other words, the code doesn't take "visibility" into account. It simply sets a date range. 


```suggestion
### Export Appointment Occurrences within the Specified Date Interval
```

concepts/05 UI Components/Scheduler/035 Occurrences.md
```diff
@@ -0,0 +1,46 @@
...
+
+Use this approach when you export only the visible appointments from the Scheduler. The mapped array converts occurrences to a flat structure suitable for CSV, Excel, or API payloads.
```
Abadzhev:
```suggestion
Use the following code snippet when you need to export events that fall within the specified date/time interval. The mapped array converts occurrences to a flat structure suitable for CSV, Excel, or API payloads.
```

concepts/05 UI Components/Scheduler/035 Occurrences.md
vladaskorohodova: I think if it's obvious, let's remove the note

PR #8674, opened 2026-04-17
Chat: Add suggestions API Description

arman-boyakhchyan:


r-farkhutdinov (approved):

api-reference/10 UI Components/dxChat/1 Configuration/suggestions.md
```diff
@@ -4,7 +4,130 @@ type: dxButtonGroup_Options
...
-<!-- Description goes here -->
+Configures suggestion buttons (DevExtreme [ButtonGroup](/Documentation/Guide/UI_Components/ButtonGroup/Getting_Started_with_ButtonGroup/)) displayed above the Chat input field.
+
+---
+Use this object to specify predefined messages or execute actions outside of the Chat. To implement suggestions, configure the following [ButtonGroup properties](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/):
+
+- The [items[]](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/items/) array
+- An [onItemClick](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/#onItemClick) handler
+
+Chat ignores the following ButtonGroup options in the **suggestions** object:
```
Abadzhev: When we have simple words like "grid, chat, scheduler" - I don't want to use them as proper nouns. It's not really the name in the API (most of the time) and it sounds vague (what chat?). 

If it was `DxDataGrid`, then okay - it's clear we refer to a type. 

Or if it was `gridControl1`, then also fine - we're referring to a specific instance. 

But simply "Grid" is strange.

```suggestion
The Chat component ignores the following ButtonGroup options in the **suggestions** object:
```

api-reference/10 UI Components/dxChat/1 Configuration/suggestions.md
```diff
@@ -4,7 +4,130 @@ type: dxButtonGroup_Options
...
+
+- [selectionMode](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/#selectionMode)
+- [selectedItemKeys](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/#selectedItemKeys)
+- [selectedItems](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/#selectedItems)
+- [onSelectionChanged](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/#onSelectionChanged)
+
+Chat overrides the default values of the following ButtonGroup properties:
```
Abadzhev:
```suggestion
The Chat component overrides the default values of the following ButtonGroup properties:
```

api-reference/10 UI Components/dxChat/1 Configuration/suggestions.md
```diff
@@ -4,7 +4,130 @@ type: dxButtonGroup_Options
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Configures suggestion buttons (DevExtreme [ButtonGroup](/Documentation/Guide/UI_Components/ButtonGroup/Getting_Started_with_ButtonGroup/)) displayed above the Chat input field.
+
+---
+Use this object to specify predefined messages or execute actions outside of the Chat. To implement suggestions, configure the following [ButtonGroup properties](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/):
+
+- The [items[]](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/items/) array
```
Abadzhev, line 12: Do I understand it correctly that button captions and text pasted into the message input are always the same? Can I have a shorter button caption ("Weather") and a longer prompt pasted into the input ("Give me a weather report for the following week and specifically focus on changes of precipitation.") ?

That seems like it should be a common task, but I don't see if that's possible.

arman-boyakhchyan: Yes, that is possible. To achieve this, customers need to specify a separate custom field for the message in the items[] array (in addition to `text`). Then, in the onItemClick handler, this custom message field would need to be used instead of `text`. We have even implemented this approach in an upcoming DevExtreme Chat [demo](https://az-jsserver.corp.devexpress.com:82/jQuery/Demos-26.1/WidgetsGallery/Demo/Chat/PromptSuggestions/FluentBlueLight/).

api-reference/10 UI Components/dxChat/1 Configuration/suggestions.md
```diff
@@ -4,7 +4,139 @@ type: dxButtonGroup_Options
...
+- An [onItemClick](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/#onItemClick) handler
+
+You can store message texts in the **items[]**.[text](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/items/#text) field. This approach specifies identical values for message texts and suggestion button captions. To use different texts, specify message texts in a custom field:
```
Abadzhev: I'm not sure why you say "a custom field" (instead of "the `message` field"). Do you mean that this value will not be automatically used as message text? I'm guessing I should write code that obtains that message and pastes it to the message box? 

Does the following change sound right? 

```suggestion
If you only specify the [text](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/items/#text) field for array items, the same text is used as the button caption and the corresponding chat message. In many cases you would prefer to use two strings: a shorter button caption and an expanded message. To specify a separate message text, assign it to a custom field and then obtain that message in the `onItemClick` handler:
```

arman-boyakhchyan: As part of our suggestions feature, no specific field is automatically used within actions (`onItemClick` handlers). Customers must manually configure which item field is used when a suggestion is clicked, whether they choose to use `text` or something else. As for `message`, this is a custom field because **dxButtonGroup**.[items[]](https://js.devexpress.com/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/items/) does not contain such a field.

To add clarity to this paragraph, it is more accurate to say that specifying message texts in `text` makes message texts and suggestion captions identical. Furthermore, rather than instructing customers to specify message texts in a custom field, we can emphasize the creation of a custom field itself. What do you think?

```suggestion
You can store message texts in the **items[]**.[text](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/items/#text) field. If you use this approach, message texts and suggestion captions are identical. To specify different texts, define a custom field for message texts:
```

Abadzhev: We are kind of returning back to a previous version. You just said that regardless of options, the message text must be specified manually in the handler. So now the second sentence seems misleading ("if you use this approach, message and caption are identical"). They are only identical, provided I write certain code in the handler. So how about this version that - that should consolidate all those details:

```suggestion
The [text](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/items/#text) field in an array item specifies the corresponding button caption. When handling the `onItemClick` event, you can choose to use the same text as a chat message. But in many cases you would prefer to use two strings: a shorter button caption and an extended message. To specify a separate message text, assign it to a custom field:
```

arman-boyakhchyan: Yeah, that works, thank you!

api-reference/10 UI Components/dxChat/1 Configuration/suggestions.md
```diff
@@ -4,7 +4,139 @@ type: dxButtonGroup_Options
 ---
 ---
 ##### shortDescription
-<!-- Description goes here -->
+Configures suggestion buttons (DevExtreme [ButtonGroup](/Documentation/Guide/UI_Components/ButtonGroup/Getting_Started_with_ButtonGroup/)) displayed above the Chat input field.
+
+---
+Use this object to specify predefined messages or execute actions outside of the Chat. To implement suggestions, configure the following [ButtonGroup properties](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/):
```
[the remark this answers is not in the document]
dirkpieterse:
```suggestion
Use this object to specify predefined messages or execute actions outside the Chat. Configure the following [ButtonGroup properties](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/) to implement suggestions:
```

api-reference/10 UI Components/dxChat/1 Configuration/suggestions.md
```diff
@@ -4,7 +4,146 @@ type: dxButtonGroup_Options
...
 
 ---
-<!-- Description goes here -->
\ No newline at end of file
+Use this object to specify predefined messages or execute actions outside of the Chat. To implement suggestions, configure the following [ButtonGroup properties](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/):
+
+- The [items[]](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/items/) array
+- An [onItemClick](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/#onItemClick) handler
+
+The [text](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/items/#text) field in an array item specifies the corresponding button caption. When handling the `onItemClick` event, you can choose to use the same text as a chat message. But in many cases you would prefer to use two strings: a shorter button caption and an extended message. To specify a separate message text, assign it to a custom field:
```
dirkpieterse:
```suggestion
The [text](/Documentation/ApiReference/UI_Components/dxButtonGroup/Configuration/items/#text) field in an array item specifies the corresponding button caption. In the `onItemClick` event handler, you can use the same text as a chat message. In many cases, however, it is better to use two strings: a shorter button caption and a longer message. To specify a separate message text, assign it to a custom field:
```

PR #8738, opened 2026-05-07
Node.js: commit the latest version

pharret31:


.github/actions/setup-env/action.yml
```diff
@@ -12,7 +12,7 @@ runs:
     - name: Setup Node.js
       uses: actions/setup-node@v4
       with:
-        node-version: 20.x
+        node-version-file: ${{ inputs.path }}/.node-version
```
dmlvr, line 16: Is this line correct?

`node-version-file: ${{ inputs.path }}/.node-version`

dmlvr (changes requested):

.github/actions/setup-env/action.yml
pharret31: Yep

dmlvr (approved):

.github/actions/setup-env/action.yml
dmlvr: resolved

PR #9086, opened 2026-08-07
HtmlEditor: Add toolbar.items[].widget Auto-Imported Modules Info

arman-boyakhchyan:


16adianay (approved):

api-reference/_hidden/dxHtmlEditorToolbarItem/widget.md
arminal, line 3: Oddly worded. How about this:
A DevExtreme component that can be used as a toolbar item.
Maybe?

arman-boyakhchyan: I believe your suggestion no longer conveys the fact that this value sets the item's componentI propose we simplify & simultaneously add info to this description as follows:
> The item component. Review [ToolbarItemComponent](https://js.devexpress.com/Documentation/ApiReference/Common_Types/#ToolbarItemComponent) for a list of supported DevExtreme components.

What do you think?

arminal: Yes

PR #9119, opened 2026-08-19
Fluent Next Theme: Add New Topics & Old Topic Updates

arman-boyakhchyan:


concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/05 Variable Naming Convention.md
```diff
@@ -0,0 +1,9 @@
+The following naming convention applies to CSS variables in Fluent Next themes:
+
+- `--dxds-*`    
+Design System CSS variables that serve as public APIs. Use these variables to customize your DevExtreme-powered application.
```
pharret31, line 4: Actually these variables may be changed as well?

dmlvr (approved):

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/05 Variable Naming Convention.md
arman-boyakhchyan: Ah, I forgot to let you guys know that I am aware that some variables are going to change. I am going to be checking & updating all names here and in the code snippets once the updated naming is finalized.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/10 Accent Colors/05 Predefined Accent Colors.md
```diff
@@ -0,0 +1,1541 @@
+Fluent Next themes ship with 11 predefined accent colors. The blue accent color is available as part of complete theme stylesheets (for instance, `dx.fluent-next.blue.light`). To apply another color, add one of the following `:root` styles to your application before you load a Fluent Next stylesheet:
```
EugeniyKiyashko: The instruction to add the :root block before the Fluent Next stylesheet inverts the cascade. dx.fluent-next.blue.light.css declares the whole accent ramp inside its single :root block, so a user :root block has identical specificity and wins only if it comes after the theme stylesheet. As written, the override silently has no effect. It also contradicts 10 Custom Accent Colors.md, which says "load after the theme stylesheet".

concepts/60 Themes and Styles/00 Styling Overview/05 Themes.md
```diff
@@ -0,0 +1,14 @@
+DevExtreme ships with a set of built-in themes:
+
+- **Fluent Next**    
+[Design System](https://docs.devexpress.com/DesignSystem/405636/foundation)-based themes that use CSS variables to declare color, typography, spacing, border, opacity, and shadow styles. Key features include:
```
EugeniyKiyashko, line 4:
```suggestion
[Design System](https://docs.devexpress.com/DesignSystem/405636/foundation) - based themes that use CSS variables to declare color, typography, spacing, border, opacity, and shadow styles. Key features include:
```

concepts/60 Themes and Styles/00 Styling Overview/05 Themes.md
```diff
@@ -0,0 +1,14 @@
...
+
+    - Predefined accent colors
+    - Custom accent color support
+    - Light and dark modes
+    - Compact mode
+    - High contrast mode
```
EugeniyKiyashko: "High contrast mode" isn't shipping for Fluent Next in 26.2. theme-options.cjs builds only light/dark × default/compact (4 bundles: dx.fluent-next.blue.{light,dark}[.compact].css), the generated token tree contains only semantic/colors/light.scss and dark.scss, and the built CSS has zero forced-colors / prefers-contrast / -ms-high-contrast rules. Please remove the bullet or mark it as planned.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/00 CSS Variables.md
```diff
@@ -0,0 +1,72 @@
...
+    <!-- tab: CSS -->
+    /* Using dxds variables */
+    .dark-colors-dx {
+        --dxds-color-surface-neutral-default-rest: var(--dxds-color-surface-neutral-default-inverted-rest);
+        --dxds-color-content-neutral-default-rest: var(--dxds-color-content-neutral-default-inverted-rest);
+        --dxds-color-border-neutral-default-rest: var(--dxds-color-border-neutral-default-inverted-rest);
+    }
+
+    /* Using custom colors */
+    .dark-colors-custom {
```
EugeniyKiyashko: .dark-colors-custom sets the surface and the content color to the same #341A51, so text is invisible on its own background. Suggest a light content value, e.g. --dxds-color-content-neutral-default-rest: #F5F0FA;.

concepts/60 Themes and Styles/05 Predefined Themes/00 Predefined Themes.md
```diff
@@ -1,71 +1,50 @@
...
-    </div>
-    <div>
-        <p>Soft Blue Compact</p>
-        <img src="/images/themes/generic-softblue-compact.png" />
-    </div>
+The following colors are available:
+
+## Fluent Next Themes
+
+TBA: Need to discuss images in this section w/ designer
```
EugeniyKiyashko: TBA: Need to discuss images in this section w/ designer would ship to the public docs. Fluent Next is also the only theme block on this page without images, while it's now listed first.

concepts/60 Themes and Styles/05 Predefined Themes/00 Predefined Themes.md
```diff
@@ -1,71 +1,50 @@
-DevExtreme provides Generic, Generic Compact, Material Design, and Fluent (all trademarks or registered trademarks are property of their respective owners) themes. These themes are available in the following colors:
+DevExtreme ships with the following themes (all trademarks or registered trademarks are property of their respective owners):
```
EugeniyKiyashko, line 1: 05 Predefined Themes/00 Predefined Themes.md:306 (+ the rename of 00 Styling Methods.md → 02 Styling Methods.md)

The rename leaves 9 topics pointing at the old file path /concepts/60%20Themes%20and%20Styles/00%20Styling%20Methods.md. Verified against the base branch - line 306 of this file plus:

concepts/60 Themes and Styles/10 HTML-Based Components Customization/HTML-Based Components Customization.md
concepts/40 Angular Components/10 Getting Started/03 Add DevExtreme to an Angular CLI Application/03 Configure Stylesheets.md
concepts/40 Angular Components/10 Getting Started/30 Other Approaches/{04 Using Webpack/04,05 Using Ionic/05,06 Using Rollup/04} Import Stylesheets.md
concepts/50 React Components/05 Add DevExtreme to a React Application/40 Import Stylesheets.md
concepts/55 Vue Components/05 Add DevExtreme to a Vue Application/40 Import Stylesheets.md
concepts/Common/Security Considerations/20 HTML Encoding/30 Potentially Vulnerable API/customizeTooltip.md

Please update all of them (or keep the original file name).

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/10 Accent Colors/05 Predefined Accent Colors.md
EugeniyKiyashko: Accent colors recolor HTML-based components only. js/__internal/viz/core/themes/fluent-next/index.ts registers the Fluent Next viz themes as plain aliases of fluent.blue.light/fluent.blue.dark, so charts and other SVG-based components stay blue no matter which accent is applied, and they don't read --dxds-* at all. This needs an explicit note in this section - otherwise a user switching to Rose will report it as a bug.

concepts/60 Themes and Styles/00 Styling Overview/05 Themes.md
```diff
@@ -0,0 +1,14 @@
...
+    - Theme customization using public `--dxds-*` CSS variables
+
+- **Fluent, Material, and Generic**    
+Themes that use SCSS to implement styles. For more information, refer to the following help topic: [Predefined Themes]().
```
EugeniyKiyashko: [Predefined Themes]() has an empty link target

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/05 Variable Naming Convention.md
```diff
@@ -0,0 +1,9 @@
...
+
+- `--dx-*`    
+Internal CSS variables and classes used by DevExtreme components. May change between versions.
```
EugeniyKiyashko: This page says --dx-* covers "internal CSS variables and classes" and recommends not using them, but 00 Styling Overview/10 CSS Styles.md recommends the opposite for classes ("You can also use these CSS classes in custom elements…") and links to the public CSS Classes API reference. Please scope the recommendation to variables, so the two topics don't contradict each other.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/10 Accent Colors/10 Custom Accent Colors.md
```diff
@@ -0,0 +1,29 @@
+You can define a custom accent color as a CSS variable and use this color to calculate `--dxds-primary-*` variables as follows:
+
+    <!-- tab: Custom -->
+    /* Set --dx-accent-color to a brand color; load after the theme stylesheet. */
+    :root {
```
EugeniyKiyashko, line 4: With the documented knobs the recipe doesn't reproduce the designed palettes. Feeding the recipe its own default #0f6cbd and comparing against _design-system/fluent/accents/blue.scss, the light half is off by ΔE(OKLab) 3–6 - step 10 comes out L 0.950 / C 0.040 vs designed L 0.977 / C 0.007. Those steps back real roles (--dxds-color-surface-primary-deep-* and …-subdued-*, i.e. hover/selected backgrounds), so the difference is visible. --dx-accent-lightness-max: 0.975 + --dx-accent-chroma-min: 0.01 cut the worst light-half deviation from 5.86 to 3.04. Worth stating too that a derived ramp is an approximation of the designed ones.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/10 Accent Colors/10 Custom Accent Colors.md
```diff
@@ -0,0 +1,29 @@
+You can define a custom accent color as a CSS variable and use this color to calculate `--dxds-primary-*` variables as follows:
+
+    <!-- tab: Custom -->
+    /* Set --dx-accent-color to a brand color; load after the theme stylesheet. */
```
EugeniyKiyashko: The snippet asks users to define --dx-accent-color and --dx-accent-*, but 05 Variable Naming Convention.md calls --dx-* internal and advises against using it. These variables also aren't declared by the theme - the user introduces them - so a neutral prefix (--my-accent-color, --my-accent-lightness-max, …) removes the contradiction and avoids a future clash with the component tier.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/10 Accent Colors/10 Custom Accent Colors.md
```diff
@@ -0,0 +1,29 @@
...
+    :root {
+        --dx-accent-color-source: var(--dx-accent-color, #0f6cbd);
+        --dx-accent-lightness-max: 0.95;
+        --dx-accent-lightness-min: 0.15;
+        --dx-accent-chroma-min: 0.04;
+
+        --dxds-primary-10: oklch(from var(--dx-accent-color-source) calc(l + 9 * (var(--dx-accent-lightness-max) - min(l, var(--dx-accent-lightness-max))) / 9) calc(c - 9 * (max(c, var(--dx-accent-chroma-min)) - var(--dx-accent-chroma-min)) / 9) h);
```
EugeniyKiyashko: oklch(from …) (relative color syntax) has no fallback here. Where it isn't supported, every --dxds-primary-* becomes invalid at computed-value time, so primary surfaces don't degrade gracefully - they lose their color completely. Worth one sentence on the browser requirement (kindly check browsers that devextreme supports, leave the note if need), and optionally wrapping the block in @supports (color: oklch(from red l c h)) so unsupported browsers keep the bundled ramp.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/10 Accent Colors/05 Predefined Accent Colors.md
EugeniyKiyashko: Two simplifications worth considering: (a) 26.2 already builds ready-to-use accent stylesheets - artifacts/css/accents/<accent>.css → dist/css/accents/ ([DevExpress/DevExtreme#34767](https://github.com/DevExpress/DevExtreme/issues/34767)) - so the recipe could be "link this file after the theme" instead of pasting 136 declarations; (b) all 11 palettes are identical except the --dxds-primary-* ramp - I compared all 136 tokens across all 11 files - so the code blocks could be trimmed to the 18 primary steps. As a bonus, the primitives are byte-identical between the light and dark bundles, so one block legitimately covers both modes; that's worth stating explicitly.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/00 CSS Variables.md
```diff
@@ -0,0 +1,72 @@
...
+        /* Shadow */
+        box-shadow: var(--dxds-box-shadow-sm);
+    }
+
+    .info-card:hover {
+        background-color: var(--dxds-color-surface-neutral-default-hovered);
+        border-color: var(--dxds-color-border-neutral-default-hovered);
+    }
+
+You can also override variables to modify styles for DevExtreme components. This allows you to apply unique styles to different parts of your application. For instance, you can use the following [semantic variable](https://docs.devexpress.com/DesignSystem/405706/colors/color-css-variables) overrides to apply dark mode styles to parts of a light mode application:
```
EugeniyKiyashko: Container-scoped overrides don't reach overlays: popups, dropdowns, toasts and drag previews render at viewport level, outside .dark-colors-dx, so they inherit the :root values - a SelectBox inside the container gets a light dropdown. Legacy swatches solved this: core/utils/swatch_container.ts finds the nearest dx-swatch-* ancestor and recreates a container with that class inside the viewport. The mechanism keys off the class-name prefix only (no compiled swatch CSS required), so naming the scope class dx-swatch-mydark makes the variable overrides follow overlays. Worth documenting here, since 55 Color Swatches.md now redirects swatch users to this section.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/00 CSS Variables.md
```diff
@@ -0,0 +1,72 @@
+Fluent Next themes ship with CSS variables defined in the DevExpress Design System. Refer to the following help topics for additional information about key styling principles and a list of available CSS variables:
+
+- [Colors](https://docs.devexpress.com/DesignSystem/405706/colors/color-css-variables)
```
EugeniyKiyashko: The "Colors" link goes to the DS semantic variables page, which documents the --dxds-color-* roles but not the --dxds-primary-NN primitives that the Accent Colors topics rely on. Consider linking the DS palette/primitives page as well.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/00 CSS Variables.md
```diff
@@ -0,0 +1,72 @@
...
+    .info-card:hover {
+        background-color: var(--dxds-color-surface-neutral-default-hovered);
+        border-color: var(--dxds-color-border-neutral-default-hovered);
+    }
+
+You can also override variables to modify styles for DevExtreme components. This allows you to apply unique styles to different parts of your application. For instance, you can use the following [semantic variable](https://docs.devexpress.com/DesignSystem/405706/colors/color-css-variables) overrides to apply dark mode styles to parts of a light mode application:
+
+    <!-- tab: CSS -->
+    /* Using dxds variables */
+    .dark-colors-dx {
```
EugeniyKiyashko: [minor] .dark-colors-dx is reused for three different examples, and the yellow accent example isn't about dark colors. Distinct class names would read better.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/10 Accent Colors/05 Predefined Accent Colors.md
```diff
@@ -0,0 +1,1541 @@
...
+
+    <!-- tab: Blue -->
+    :root {
+        --dxds-primary-10: #f4f8fc;
+        --dxds-primary-20: #e7eff9;
+        --dxds-primary-30: #cce1f8;
+        --dxds-primary-40: #b4d2f4;
+        --dxds-primary-50: #9cc3f0;
+        --dxds-primary-60: #82b4ea;
+        --dxds-primary-70: #67a2e1;
```
EugeniyKiyashko: [minor] 9 * (…) / 9 in the step-10 formula is a no-op; --dxds-primary-100: oklch(from … l c h) could simply be the source color. Simplifying makes the pattern easier to read.

concepts/60 Themes and Styles/00 Styling Overview/10 CSS Styles.md
```diff
@@ -0,0 +1,3 @@
+You can apply custom CSS styles to specific classes to customize DevExtreme component appearance. You can also use these CSS classes in custom elements to ensure a consistent look and feel across your app.
+
+For a list of common DevExtreme classes, refer to the following help topic: [CSS Classes](/Documentation/ApiReference/UI_Components/CSS_Classes/).
```
EugeniyKiyashko: **Missing trailing newline in** 00 Styling Overview/10 CSS Styles.md, 07 Fluent Next Theme Customization/00 Fluent Next Theme Customization.md, 05 CSS Variables/05 Variable Naming Convention.md, 10 Accent Colors/00 Accent Colors.md.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/10 Accent Colors/05 Predefined Accent Colors.md
```diff
@@ -0,0 +1,1541 @@
...
+        --dxds-neutral-220: #3b3b3b;
+        --dxds-neutral-230: #333333;
+        --dxds-neutral-240: #2c2c2c;
+        --dxds-neutral-250: #242424;
+        --dxds-neutral-260: #1d1d1d;
+        --dxds-neutral-270: #161616;
+        --dxds-neutral-280: #0a0a0a;
+    }
+
+    <!-- tab: Cool Blue -->
```
EugeniyKiyashko, line 96: [suggestion]

The product already ships ready-to-use accent stylesheets, so this topic doesn't need 1500 lines of pasted declarations. scss-build emits artifacts/css/accents/<accent>.css and the whole dist/css tree is published with the devextreme package, so the recipe can be a single link/import after the theme stylesheet:

<link rel="stylesheet" href="node_modules/devextreme/dist/css/dx.fluent-next.blue.light.css">
<link rel="stylesheet" href="node_modules/devextreme/dist/css/accents/rose.css">
import 'devextreme/dist/css/dx.fluent-next.blue.light.css';
import 'devextreme/dist/css/accents/rose.css';

Available files: blue, cool-blue, desert, mint, moss, orchid, purple, rose, rust, steel, storm (blue is already embedded in the theme bundles, so linking it is a no-op). Three details worth stating alongside:

Order matters - the file declares the ramp on :root, exactly like the theme bundle, so it only wins if it comes after the theme (same issue as the "before you load" wording in the intro).
One file covers every mode and size - the primitives are byte-identical in the light, dark and compact bundles, so switching modes means swapping only the theme file, not the accent file.
The file pins all 136 primitives, not just --dxds-primary-* (the 11 palettes differ only in the primary ramp). Harmless, but a user's own primitive overrides must be loaded after it.

Suggested structure: a table of the 11 names + the two-line snippet above, and keep the full variable dumps only as an optional "inline the values instead" sample - trimmed to the 18 --dxds-primary-* steps, since the rest is identical across all accents. Note that a file-based accent exists only for these 11 designed palettes; an arbitrary brand color remains the CSS recipe in 10 Custom Accent Colors.md.

concepts/60 Themes and Styles/00 Styling Overview/05 Themes.md
arman-boyakhchyan: This is not a typо, this text will be displayed as "Design Sytem-based themes" on the site, worded identically to our Blazor Fluent's description on https://docs.devexpress.com/Blazor/404821/styling-and-themes/styling-and-themes#themes

concepts/60 Themes and Styles/05 Predefined Themes/00 Predefined Themes.md
arman-boyakhchyan: Added images from regular fluent for the time being, updating the design & images of this section (I believe it looks a bit strange) will be a task for a later time.

concepts/60 Themes and Styles/05 Predefined Themes/00 Predefined Themes.md
arman-boyakhchyan: I have renamed Styling Methods to add the new Styling Overview topic at the top of this section within the site's left-hand menu. Renamings like these, where only the number at the front of the file name is changed, do not break links. Our docgen mechanism will update 00 > 02 automatically, we do not add such updates by hand. Link updates are only necessary when text changes.

concepts/60 Themes and Styles/05 Predefined Themes/00 Predefined Themes.md
```diff
@@ -1,71 +1,69 @@
...
-    </div>
-    <div>
-        <p>Dark Compact</p>
-        <img src="/images/themes/generic-dark-compact.png" />
-    </div>
+The following colors are available out of the box:
+
+## Fluent Next Themes
+
+<div id="fluent-next-themes" class="theme-images">
```
EugeniyKiyashko, line 12: The Fluent Next block reuses the Fluent screenshots (fluent-default.png, fluent-compact.png, fluent-dark-default.png, fluent-dark-compact.png) - the same four images as the Fluent section below, so both themes look identical on this page. Are the Fluent Next captures still pending?

concepts/60 Themes and Styles/05 Predefined Themes/60 Switch Between Themes at Runtime/2 Without Page Reload.md
```diff
@@ -110,6 +110,8 @@ You can use this approach only if the themes belong to the same group. For insta
 
 The method accepts the `data-theme` attribute value. For example, if you want to switch to `dx.generic.dark`, pass `generic.dark` (without the 'dx' prefix) as an argument in the method. For a custom theme, specify the theme you used as a base. If both of your themes are derived from the same base, get the `data-theme` value from the `font-family` value of the `dx-theme-marker` CSS class, which is located in each theme's CSS file.
 
+[note] If you use an accent color stylesheet with Fluent Next themes, you do not need to switch the accent stylesheet when switching themes. These stylesheets apply to all variations of Fluent Next (light and dark modes, standard and compact sizes). Refer to the following help topic for more information about accent colors in Fluent Next themes: [Fluent Next Theme Customization - Accent Colors](/Documentation/Guide/Themes_and_Styles/Fluent_Next_Theme_Customization/#Accent_Colors).
```
EugeniyKiyashko: Correct, with one condition worth spelling out: DevExtreme replaces all rel="dx-theme" links with a single <link rel=stylesheet> inserted at the position of the last dx-theme link and then only swaps its href ($allThemeLinks.last().after($activeThemeLink); $allThemeLinks.remove(); in themes.ts). So the accent stylesheet keeps working across switches only if its <link> comes after the whole dx-theme block - placed before or among them, it silently loses the cascade. Worth adding to the note and showing in the <head> snippet.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/10 Accent Colors/05 Predefined Accent Colors.md
```diff
@@ -0,0 +1,314 @@
+Fluent Next themes ship with 11 predefined accent colors. To apply one of these colors, import an accent stylesheet after the theme:
+
+    <!-- tab: Modular Application -->
+    import 'devextreme/dist/css/dx.fluent-next.blue.light.css';
+    import 'devextreme/dist/css/accents/rose.css';
+
+    <!-- tab: Non-Modular Application -->
+    <link rel="stylesheet" href="node_modules/devextreme/dist/css/dx.fluent-next.blue.light.css"></link>
+    <link rel="stylesheet" href="node_modules/devextreme/dist/css/accents/rose.css"></link>
```
EugeniyKiyashko: [validate it] <link> is a void element - drop the </link> closing tags in the non-modular sample.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/05 Variable Naming Convention.md
EugeniyKiyashko: The recommendation below now correctly mentions variables only, but this bullet still reads "Internal CSS variables and classes", which contradicts 00 Styling Overview/10 CSS Styles.md (it recommends using DevExtreme CSS classes and links to the public CSS Classes reference). Suggest dropping "and classes" here.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/00 CSS Variables.md
```diff
@@ -0,0 +1,213 @@
+Fluent Next themes ship with CSS variables defined in the DevExpress Design System. Refer to the following help topics for additional information about key styling principles and a list of available CSS variables:
```
EugeniyKiyashko, line 1: Heads-up: the base branch moved to design tokens 262.10.1 on Aug 26 (07fad6c7f8), which renamed the semantic color tokens to a shorter grammar, and the old names are gone from the product - zero occurrences of color-surface-, -default-rest, color-content-neutral or utility-yellow remain in packages/devextreme-scss/scss and apps/demos/Demos, and there are no aliases (the demos were rewritten in b2f401141e). Every semantic color variable in this topic is a pre-262.10.1 name, so the snippets will silently do nothing (an unknown custom property is invalid at computed-value time and resolves to unset). Mapping for the names used here:

--dxds-color-surface-neutral-default-rest → --dxds-color-bg
--dxds-color-surface-neutral-default-hovered → --dxds-color-bg-hovered
--dxds-color-surface-neutral-default-inverted-rest → --dxds-color-bg-inverted
--dxds-color-content-neutral-default-rest → --dxds-color-content
--dxds-color-content-neutral-default-inverted-rest → --dxds-color-content-inverted
--dxds-color-border-neutral-default-rest → --dxds-color-border
--dxds-color-border-neutral-default-hovered → --dxds-color-border-hovered
--dxds-color-border-neutral-default-inverted-rest → --dxds-color-border-inverted (pattern-based; the theme doesn't consume it, so please confirm against the package)
--dxds-color-surface-primary-default-rest → --dxds-color-bg-primary
--dxds-color-content-primary-default-rest → --dxds-color-content-primary
--dxds-color-border-primary-default-rest → --dxds-color-border-primary
--dxds-utility-yellow-40/50/60 - needs a decision: utility-* no longer appears anywhere in the product, and the semantic equivalents are now --dxds-color-bg-yellow, --dxds-color-bg-yellow-subtle, --dxds-color-content-yellow. Using the semantic yellow roles here would also be more consistent with the rest of the topic.

The scales in the .info-card sample (--dxds-spacing-*, --dxds-border-width-10, --dxds-border-radius-40, --dxds-font-size-base-md, --dxds-line-height-base-md, --dxds-box-shadow-sm) and the whole Accent Colors section (--dxds-primary-10…180) are unaffected - no changes needed there.

One more coordination point: the DS pages linked at the top still document the old grammar (405706 has 308 --dxds-color-surface-* references and no --dxds-color-bg; 405687 documents --dxds-utility-*), so readers who follow those links will get names that don't exist in 26.2. Worth syncing with the Design System doc update before publishing.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/00 Fluent Next Theme Customization.md
```diff
@@ -0,0 +1 @@
+Fluent Next themes are based on the DevExpress Design System and support multiple customization options.
```
alexfi1in: Could we specify that Fluent Next themes are based on the Design System Foundation and add a link to the corresponding documentation?

Suggested wording:

Fluent Next themes are based on the [Design System Foundation](https://docs.devexpress.com/DesignSystem/405636/foundation) and support multiple customization options.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/10 Accent Colors/05 Predefined Accent Colors.md
```diff
@@ -0,0 +1,314 @@
+Fluent Next themes ship with 11 predefined accent colors. To apply one of these colors, import an accent stylesheet after the theme:
```
alexfi1in: These predefined colors are part of the DevExpress Design System Foundation. Could we link to the source palettes and clarify the relationship between the Design System's primary color palettes and DevExtreme accent colors?

Suggested wording:

The DevExpress Design System defines [11 predefined primary color palettes](https://docs.devexpress.com/DesignSystem/405638/colors/theme-palettes/fluent-theme-palettes) for Fluent themes. To use one of these palettes as the accent color in a Fluent Next theme, import the corresponding accent stylesheet after the theme stylesheet:

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/00 CSS Variables.md
```diff
@@ -0,0 +1,213 @@
...
+[/note]
+
+You can also use CSS variable overrides to apply custom colors to specific parts of your application. You can define custom colors or use [utility palette](https://docs.devexpress.com/DesignSystem/405639/colors/utility-palettes/fluent-utility-palette) colors:
+
+    <!-- tab: CSS -->
+    /* Utility palette colors */
+    .yellow-accent {
+        --dxds-color-bg: var(--dxds-color-bg-yellow);
+        --dxds-color-content: var(--dxds-color-content-yellow);
+        --dxds-color-border: var(--dxds-color-bg-yellow-subtle);
```
EugeniyKiyashko:
```suggestion
        --dxds-color-border: var(--dxds-color-border-yellow);
```

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/15 Theme Modes/05 Container-Specific Theme Modes.md
```diff
@@ -0,0 +1,166 @@
...
+                .removeClass('dx-theme-mode-light')
+                .addClass('dx-theme-mode-dark');
+            DevExpress.ui.themes.refreshMode();
+        },
+    })
+
+##### Angular
+
+    <!-- tab: app.component.ts -->
+    import refreshMode from "devextreme/ui/themes";
```
EugeniyKiyashko, line 41: This is a default import: it binds the whole `themes` object to the name `refreshMode`, so the call below is not a function. Running tsc on the snippet as written:

```
error TS2348: Value of type 'typeof themes' is not callable.
```

It needs to be a named import:

```ts
import { refreshMode } from "devextreme/ui/themes";
```

The same line appears in the Vue snippet (line 76), the React one (line 92), and all three `mode()` samples further down (lines 143, 154, 162).

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/15 Theme Modes/05 Container-Specific Theme Modes.md
```diff
@@ -0,0 +1,166 @@
...
+                ></dx-button>
+            </div>
+        `,
+    })
+    export class AppComponent {
+        containerClass: string = 'dx-theme-mode-light';
+
+        changeThemeMode() {
+            this.containerClass = 'dx-theme-mode-dark';
+            refreshTheme();
```
EugeniyKiyashko, line 59: refreshTheme() does not exist. The function is refreshMode() - the React snippet below calls it correctly.

The Vue snippet has the same mistake on line 83.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/15 Theme Modes/05 Container-Specific Theme Modes.md
```diff
@@ -0,0 +1,166 @@
...
+                @click="changeThemeMode"
+            />
+        </div>
+    </template>
+
+    <script setup lang="ts">
+    import refreshMode from "devextreme/ui/themes";
+    import { DxButton } from 'devextreme-vue/button';
+
+    const containerClass: string = 'dx-theme-mode-light';
```
EugeniyKiyashko, line 79: Two more problems here, both independent of the import:

- containerClass is declared const and then reassigned on line 82, which throws.
- Even without that, a plain variable in `<script setup>` will not re-render the template - it has to be a ref.

```ts
import { ref } from "vue";
import { refreshMode } from "devextreme/ui/themes";

const containerClass = ref("dx-theme-mode-light");

function changeThemeMode() {
    containerClass.value = "dx-theme-mode-dark";
    refreshMode();
}
```

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/15 Theme Modes/05 Container-Specific Theme Modes.md
```diff
@@ -0,0 +1,166 @@
...
+    </body>
+
+You can call the [mode(element)]() method to get the calculated theme mode of an element (*"light"* or *"dark"*). The following code snippet returns *"light"* for the `.three` element from the previous sample:
+
+---
+
+##### jQuery  
+
+    <!-- tab: index.js -->
+    const calculatedMode = DevExpress.ui.themes.mode('.three');
```
EugeniyKiyashko, line 138: mode() takes an element, not a selector. With the import corrected, the snippet still fails to compile:

```
error TS2345: Argument of type 'string' is not assignable to parameter of type 'Element'.
```

A string does happen to work at runtime, because the internal renderer runs querySelectorAll on it, but that is not the documented signature - ui.themes.mode is declared as Element|jQuery - so I would not show it.

jQuery:

```js
const calculatedMode = DevExpress.ui.themes.mode($(".three"));
```

TypeScript (lines 143-164):

```ts
import { mode } from "devextreme/ui/themes";
const calculatedMode = mode(document.querySelector(".three")!);
```

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/15 Theme Modes/05 Container-Specific Theme Modes.md
```diff
@@ -0,0 +1,166 @@
...
+
+[/note]
+
+Use one of the following classes to specify a theme mode for a container:
+
+- `dx-theme-mode-light`: Applies **light** mode styles to a container and its children
+- `dx-theme-mode-dark`: Applies **dark** mode styles to a container and its children
+- `dx-theme-mode-inverted`: Applies the opposite theme mode relative to a container's parent
+
+You can change container theme modes at runtime. To ensure DevExtreme component styles are updated, call [refreshMode()]() after you update a container's theme mode:
```
EugeniyKiyashko, line 16: This link has no target, and neither does [mode(element)]() on line 131.

Both methods got their @docid in DevExtreme#35011 (merged 11 Sep) and are now part of the documented API as ui.themes.refreshMode and ui.themes.mode, so the links can be filled in.

Their description texts do not exist yet - the keys are `ui.themes.mode(element)` and `ui.themes.refreshMode()`.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/10 Accent Colors/10 Custom Accent Colors.md
```diff
@@ -0,0 +1,22 @@
+Fluent Next stylesheets ship with a CSS rule that calculates [primary shades](https://docs.devexpress.com/DesignSystem/405638/colors/theme-palettes/fluent-theme-palettes#fluent-primary) from the `--dx-accent-color` variable. Assign a color to this variable to use Fluent Next themes with a custom accent color:
+
+    <!-- tab: CSS -->
+    :root {
+        --dx-accent-color: SlateBlue;
```
EugeniyKiyashko, line 5: Worth saying explicitly that --dx-accent-color only takes effect on :root.

Every example in this topic uses :root, but nothing states it is a requirement, and the natural next thing a reader tries is a different accent on one part of the page:

```css
.my-panel { --dx-accent-color: SlateBlue; }   /* silently does nothing */
```

The shades are calculated where the variable is declared, so a declaration on a descendant never reaches them. This is worth one sentence particularly because it sits next to the theme-modes topic, where scoping a class to a container is exactly what the reader is told to do.

For what it is worth, a custom accent does combine correctly with container theme modes - I measured it on the built bundle: with --dx-accent-color set on :root, an element inside a dx-theme-mode-dark container takes a different shade of the custom color than the same element outside it, which is the intended behavior.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/15 Theme Modes/05 Container-Specific Theme Modes.md
```diff
@@ -0,0 +1,166 @@
...
+    export default function App() {
+        const [containerClass, setContainerClass] = useState('dx-theme-mode-light');
+
+        const changeThemeMode = useCallback(() => {
+            setContainerClass('dx-theme-mode-dark');
+            refreshMode();
+        }, []);
+
+        return (
+            <div class={containerClass}>
```
EugeniyKiyashko, line 103: The React snippet uses the HTML attribute name. In JSX it has to be `className`, otherwise React warns and the class is never applied:

```jsx
<div className={containerClass}>
```

The other three snippets are fine on this point.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/15 Theme Modes/05 Container-Specific Theme Modes.md
EugeniyKiyashko, line 16: Two things about this sentence, both separate from the empty link.

**1. It describes the wrong mechanism, and it undersells the feature.** Components repaint on their own - that is the whole point of shipping both modes in one stylesheet, no script is involved. `refreshMode()` has exactly one subscriber in the product (`Overlay`), and it re-parents an overlay only when its container actually changed. So it matters only for an overlay that is **already open** when the class moves: a popup, drop-down or tooltip renders in the viewport, outside the container whose class you just changed. With nothing open the call is a no-op.

Suggested wording:

> You can change container theme modes at runtime; styles update on their own. If a popup, drop-down or tooltip is open at that moment, it is rendered outside the container, so call [refreshMode()] afterwards to move it into the new mode.

Worth adding that the call is safe for a focused overlay - since DevExpress/DevExtreme#35011 the focus and caret are restored after the move.

**2. The other half of the contract is missing**, here and in `05 Predefined Themes/60 Switch Between Themes at Runtime/2 Without Page Reload.md`: `refreshMode()` is also required after switching the theme itself, not only after moving a class. This is deliberate on the product side - firing it from inside `current()` was tried at two points and both read stale values, because a theme switch swaps one `<link>`'s href and the new stylesheet lands after the call returns. So the contract is one call after changing anything that affects what an element resolves to: a class, or the whole theme.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/15 Theme Modes/05 Container-Specific Theme Modes.md
```diff
@@ -0,0 +1,166 @@
+Fluent Next themes allow you to define theme modes for specific containers in your application. Container-specific modes allow you to use dark mode styles in a lght mode application and vice versa. These modes apply to DevExtreme components and HTML elements.
```
EugeniyKiyashko, line 1: Typo: "in a lght mode application" -> "light".

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/15 Theme Modes/05 Container-Specific Theme Modes.md
```diff
@@ -0,0 +1,166 @@
...
+- SVG components do not support container-specific theme modes and always use the application's theme mode (from the active Fluent Next stylesheet).
+- You do not need to import both light and dark stylesheets to use container-specific theme modes. All Fluent Next stylesheets ship with both light and dark CSS rules.
+
+[/note]
+
+Use one of the following classes to specify a theme mode for a container:
+
+- `dx-theme-mode-light`: Applies **light** mode styles to a container and its children
+- `dx-theme-mode-dark`: Applies **dark** mode styles to a container and its children
+- `dx-theme-mode-inverted`: Applies the opposite theme mode relative to a container's parent
```
EugeniyKiyashko, line 14: "relative to a container's parent" is operationally right, but it reads as if the parent has to carry a mode class itself - it does not, any ancestor scope works through inheritance. The contract wording is "the opposite of the **nearest enclosing** theme mode":

> `dx-theme-mode-inverted`: Applies the theme mode opposite to the nearest enclosing one

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/15 Theme Modes/05 Container-Specific Theme Modes.md
```diff
@@ -0,0 +1,166 @@
...
+            </div>
+        );
+    }
+
+---
+
+`dx-theme-mode-inverted` supports nesting. The following code snippet demonstrates three nesting levels:
+
+    <!-- tab: HTML -->
+    <body class="dx-viewport dx-theme-mode-dark">
```
EugeniyKiyashko, line 117: For a whole page the mode class belongs on `<html>`, not on `body`. `:root` and a class have the same specificity, so a class on `body` overrides - for everything inside it - whatever the user declared on `:root`, including their own role overrides.

Measured on the built bundle with a user override `:root { --dxds-color-bg: #ff0000 }`:

| mode class on | what an element inside sees |
|---|---|
| `<body>` | `#242424` - the override is lost |
| `<html>` | `#ff0000` - the override works |

This snippet is the first thing a reader copies for "make my app dark", so it is worth either moving the class to `<html>` or saying explicitly that this is a nesting demo and page-level modes go on `<html>`.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/15 Theme Modes/05 Container-Specific Theme Modes.md
```diff
@@ -0,0 +1,166 @@
...
+            <div class="two dx-theme-mode-inverted">
+                <!-- Dark mode within .two -->
+                <div class="three dx-theme-mode-inverted">
+                    <!-- Light mode within .three -->
+                </div>
+            </div>
+        </div>
+    </body>
+
+You can call the [mode(element)]() method to get the calculated theme mode of an element (*"light"* or *"dark"*). The following code snippet returns *"light"* for the `.three` element from the previous sample:
```
EugeniyKiyashko, line 131: Worth one sentence that `mode()` is safe to call under any theme: when nothing declares a mode it answers from the loaded stylesheet, so it returns `"light"` or `"dark"` rather than null, and it never throws - including for a detached element. As written, the topic reads as if the method only makes sense inside a mode container.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/00 CSS Variables.md
```diff
@@ -0,0 +1,213 @@
...
+                    wrapperAttr={popupWrapperAttr}
+                />
+            );
+        }
+
+    ---
+
+[/note]
+
+You can also use CSS variable overrides to apply custom colors to specific parts of your application. You can define custom colors or use [utility palette](https://docs.devexpress.com/DesignSystem/405639/colors/utility-palettes/fluent-utility-palette) colors:
```
EugeniyKiyashko, line 198: `--dx-theme-mode` is not documented anywhere in this PR. Every mode scope publishes it (`light` or `dark`), it is what `themes.mode()` reads, and it is the only way an application's own CSS can react to the mode without duplicating our logic:

```css
@container style(--dx-theme-mode: dark) {
    .my-panel { background: #1d1d1d; }
}
```

Since the mode classes are being documented as public API, this property belongs in the same contract - and this section, which is about applying overrides to specific parts of an application, is where a reader would look for it.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/15 Theme Modes/00 Theme Modes.md
```diff
@@ -0,0 +1,4 @@
+Fluent Next themes ship with light and dark theme modes. These modes are available as separate stylesheets:
+
+- `dx.fluent-next.blue.light.css`
+- `dx.fluent-next.blue.dark.css`
```
EugeniyKiyashko, line 4: The list is incomplete: Fluent Next ships four stylesheets, not two - `dx.fluent-next.blue.{light,dark}.css` plus `dx.fluent-next.blue.{light,dark}.compact.css`. Same set you quoted on `05 Themes.md` when asking to drop the high-contrast bullet.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/10 Accent Colors/05 Predefined Accent Colors.md
```diff
@@ -0,0 +1,316 @@
...
+        <td>Storm</td>
+        <td><code>accents/storm.css</code></td>
+    </tr>
+</table>
+
+[note]
+
+- Blue is the default accent color in Fluent Next themes. You do not need to import the blue accent stylesheet to apply this accent color.
+- Theme and accent stylesheets define colors on the same selector (`:root`). To ensure accent colors are applied, load your app's accent stylesheet immediately after the theme stylesheet.
+- Accent stylesheets apply to all variations of Fluent Next (light and dark modes, standard and compact sizes). To switch theme modes or sizes, swap the theme stylesheet only.
```
EugeniyKiyashko, line 74: Worth stating which accent wins when both mechanisms are used.

The generated palettes now read `--dxds-primary-NN: var(--dx-accent-color-NN, <palette hex>)` (`_design-system/fluent/accents/*.scss`), and the theme derives `--dx-accent-color-*` from `--dx-accent-color` through `--dx-accent-color-source: var(--dx-accent-color)` with no fallback (`widgets/fluent-next/_accent-color.scss`). So the palette hex is only a fallback: it applies while `--dx-accent-color` is unset, and as soon as a custom accent is set the imported palette stops having any effect - regardless of the order the stylesheets are loaded in.

The note above explains load order for theme vs accent stylesheet, and line 316 mentions the fallback form for the inline variant, but nothing tells a reader what happens when they import `rose.css` and also set `--dx-accent-color` somewhere: the page turns out in the custom color and the imported palette looks ignored.

Suggested addition to this note:

> - A custom accent color (`--dx-accent-color`) takes priority over an imported accent stylesheet, regardless of the order in which the stylesheets are loaded. Predefined shades act as fallback values and apply only while `--dx-accent-color` is not set.

(Read from the sources, not measured - the bundle in my artifacts predates the custom-accent mechanism.)

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/00 CSS Variables.md
```diff
@@ -0,0 +1,213 @@
...
+        /* Shadow */
+        box-shadow: var(--dxds-box-shadow-sm);
+    }
+
+    .info-card:hover {
+        background-color: var(--dxds-color-bg-hovered);
+        border-color: var(--dxds-color-border-hovered);
+    }
+
+CSS variable overrides also allow you to modify styles of DevExtreme components. You can define overrides for individual components or wrap multiple components in a container and define overrides on the container level. This allows you to apply unique styles to different parts of your application. The following code snippet overrides [semantic variables](https://docs.devexpress.com/DesignSystem/405706/colors/color-css-variables):
```
EugeniyKiyashko, line 51: The topic never shows the simplest case - a global override. There is not a single `:root` in this file: every example scopes to a class, so a reader who wants "recolor the whole application" has to infer the global form.

The inferred version carries a rule that is not stated anywhere in this topic: a `:root` block in a user stylesheet has the same specificity as the theme's own `:root`, so it wins only when that stylesheet is loaded after the theme. The accent topic states this rule, but only for accent stylesheets.

Suggested addition before the scoped examples:

```css
/* Applies to the whole application. Load this stylesheet after the Fluent Next theme. */
:root {
    --dxds-color-bg-primary: #b06ab3;
}
```

Worth pairing it with one sentence that the class-scoped overrides shown below work regardless of load order, because a class outranks `:root`. That is the same mechanism as in my comment on `05 Container-Specific Theme Modes.md:117`, seen from the other side: a mode class on `body` outranks a user's own `:root` overrides.

concepts/60 Themes and Styles/07 Fluent Next Theme Customization/05 CSS Variables/00 CSS Variables.md
```diff
@@ -0,0 +1,213 @@
...
+                    <Popup /> {/* Styles defined for .dx-swatch-myswatch will apply to the Popup content */}
+                </div>
+            );
+        }
+
+    ---
+
+    Note that DevExtreme does not generate `.dx-swatch-*` wrapper containers for components where the **container** property is defined.
+
+2. **Use wrapperAttr Properties**
```
EugeniyKiyashko, line 142: The two approaches cover Popup but not the more common case.

`wrapperAttr` is declared once, on the overlay base (`js/ui/overlay.d.ts:178`), so Popup, Popover, Toast and LoadPanel take it as a top-level property - which is what the snippets below show. Editors that open a drop-down (SelectBox, Lookup, DateBox, ColorBox, DropDownBox, DropDownButton, Autocomplete) expose it one level down, inside `dropDownOptions` (typed as `PopupProperties` - see `js/ui/select_box.d.ts:287`). Recoloring a drop-down list is what a reader of this section is most likely to need, and as written the topic gives no way to reach it.

Suggested addition after the Popup sample:

```js
$("#selectBox").dxSelectBox({
    dropDownOptions: {
        wrapperAttr: { class: "dark-colors-custom" }
    }
});
```

plus one sentence: "For editors that display a drop-down list, specify **wrapperAttr** within **dropDownOptions**."

